PR #5, opened 2018-08-23
*: propagate workdir to children containers

erizocosmico:
Since daemon itself was not going to do anything with the workdir decided not to mount it and just pass it as an argument. Children containers are started with the workdir mounted if needed.

Going forward containers may need some configuration and `InfoOrStart` provided too little possibility for configuration from the outside, so I added the possibility to pass it some config options.

### DIFF
--- a/cmd/srcd-server/engine/engine.go
+++ b/cmd/srcd-server/engine/engine.go
@@ -24,11 +24,13 @@ var _ api.EngineServer = new(Server)
 
 type Server struct {
 	version string
+	workdir string
 }
 
-func NewServer(version string) *Server {
+func NewServer(version, workdir string) *Server {
 	return &Server{
 		version: version,
+		workdir: workdir,
 	}
 }
 
@@ -162,6 +164,7 @@ func createBbblfshd() error {
 		Image: "bblfsh/bblfshd",
 		Cmd:   []string{"-ctl-address=0.0.0.0:9433", "-ctl-network=tcp"},
 	}
+
 	// TODO: add volume to store drivers
 	host := &container.HostConfig{Privileged: true}
 
--- a/cmd/srcd-server/engine/sql.go
+++ b/cmd/srcd-server/engine/sql.go
@@ -14,10 +14,18 @@ import (
 	"github.com/src-d/engine-cli/docker"
 )
 
-const gitbaseName = "srcd-cli-gitbase"
+const (
+	gitbaseName      = "srcd-cli-gitbase"
+	gitbaseMountPath = "/opt/repos"
+)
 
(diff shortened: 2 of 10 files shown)


cmd/srcd-server/engine/engine.go
```diff
@@ -152,7 +154,11 @@ func (s *Server) parse(ctx context.Context, req *api.ParseRequest, log logf) (*a
 	return resp, nil
 }
 
-func createBbblfshd() error {
+func (s *Server) withWorkdirMounted(at string) docker.ConfigOption {
```
campoy: wondering whether this function is actually needed
it's only called once and the body is simple enough I would just manually inline it

cmd/srcd-server/engine/sql.go
```diff
@@ -15,9 +15,14 @@ import (
 )
 
 const gitbaseName = "srcd-cli-gitbase"
+const gitbaseMountPath = "/opt/repos"
```
campoy: const (
   gitbaseName = ...
   gitbaseMountPath = ...
)

cmd/srcd/cmd/sql.go
```diff
@@ -40,7 +40,12 @@ var sqlCmd = &cobra.Command{
 		}
 		query := args[0]
 
-		c, err := daemon.Client()
+		workdir, err := os.Getwd()
+		if err != nil {
+			logrus.Fatalf("could not get working directory: %s", err)
+		}
+
+		c, err := daemon.Client(workdir)
```
campoy: I feel like you could create a new function daemon.ClientOnPwd and move the code calling os.Getwd into it

cmd/srcd/daemon/daemon.go
```diff
@@ -42,8 +43,13 @@ func Client() (api.EngineClient, error) {
 	return api.NewEngineClient(conn), nil
 }
 
-func Start() error {
+func Start(workdir string) error {
+	return start(docker.WithEnv(workdirKey, workdir))
+}
+
+func start(opts ...docker.ConfigOption) error {
```
campoy: I'd make start(opts) return a `func() error` so you don't need to pass the options around so much.

Closures ftw

erizocosmico:
Fixed

cmd/srcd/cmd/sql.go
erizocosmico: Actually, Client never needed workdir cause the only place you need a workdir different than pwd is Start, so I'm gonna remove all that

erizocosmico:
Fixed the issue. Right now it's a little hacky, because it kills all components (which is all docker containers here). I will rebase #9 with this and make it only remove the components that are workdir dependant (no need to kill bblfsh, for example, which doesn't require workdir).

PR #6, opened 2018-08-23
*: automatically pull missing images before creating containers

erizocosmico:
Depends on #5 

This contains a few changes related to srcd components in the engine:

- Now only allowed namespaces (srcd, srcd-cli and bblfsh) are allowed as components, so it no longer outputs all docker images when listing components. Also you can't install a component that's not a srcd component. We might need to figure out how to deal with dependencies (e.g. pilosa, postgres, etc) of these components. Right now there are none, so no big deal.
- Automatically pulls images that are not installed before creating the containers. This saves the user from having to manage components by hand before doing things.


erizocosmico:
Fixed

PR #7, opened 2018-08-23
cmd: output SQL as a table

erizocosmico:
Changes the output of rows from tabbed to a pretty-printed table


campoy:
UPDATE: it's so pretty!

screen shot 2018-08-23 at 1 25 31 pm

PR #8, opened 2018-08-23
cmd: interactive SQL repl

erizocosmico:
Depends on #7


cmd/srcd/cmd/sql.go
```diff
@@ -33,37 +33,75 @@ var sqlCmd = &cobra.Command{
...
-		w := new(tabwriter.Writer)
-		w.Init(os.Stdout, 0, 8, 2, '\t', 0)
-		defer w.Flush()
-
-		fmt.Fprintln(w, strings.ToUpper(strings.Join(res.Header.Cell, "\t")))
-		fmt.Fprintln(w, "-------------")
-		for _, row := range res.Rows {
-			fmt.Fprintln(w, strings.Join(row.Cell, "\t"))
+		switch strings.ToLower(strings.TrimSpace(line)) {
+		case "exit", "quit":
```
campoy: I would expect `exit;` and `quit;` to work too, but right now they fail with a syntax error.

cmd/srcd/cmd/sql.go
```diff
@@ -33,37 +33,75 @@ var sqlCmd = &cobra.Command{
...
+	if err != nil {
+		return fmt.Errorf("could not get daemon client: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	res, err := c.SQL(ctx, &api.SQLRequest{Query: query})
+	if err != nil {
+		return err
```
campoy: I'd consider adding at least at `TODO` in here regarding the error messages.
Whenever a syntax error occurs in the backend the error message looks like:

```
rpc error: code = Unknown desc = SQL query failed: Error 1105: unknown error: syntax error at position 5 near 'exit'
```

I think it'd be better to identify this was not a transport error somehow and print just the `syntax error at position 5 near 'exit'`

erizocosmico:
Done

erizocosmico:
Fixed. Did `git push` but since it was a rebase failed and didn't notice 😅

Issue #3, opened 2018-08-21
Update go imports (engine -> engine-cli)


eiso:
The name has been changed, so please update the import paths from `engine` to `engine-cli`.

PR #14, opened 2018-08-24
update import paths after repo rename

campoy:
I'll merge later on, to avoid conflicts with @erizocosmico's work


campoy:
Fixes #3

PR #26, opened 2018-08-28
daemon: use alpine as base image on second docker build stage

### LINKED ISSUE #21: srcd/cli-daemon images is 1GB

campoy:
Fixes #21.


campoy:
Image size went from 1GB to 41MB. 🎉

erizocosmico:
Does all the libuast things work in there? Last time we tried to dockerize with alpine something with libuast we couldn't

PR #53, opened 2018-09-23
Link to what is HEAD in README

vmarkovtsev:
Usually HEAD refers to the latest revision, and the phrase "commit count in HEAD" thus means "commit count in the latest commit" which makes no sense. We really mean the HEAD reference here, and I link to the description in the official documentation.


vmarkovtsev:
@smacker Yet another friendly 1 month ping :)

PR #77, opened 2018-10-15
add ';' by the end of sql queries examples

gomesfernanda:
Signed-off-by: Fernanda Gomes <fergomes@gmail.com>


smacker:
To give more context:

semicolumns are needed for command line sql interface.
currently, after copy-paste user has to add `;` manually after each query.

Issue #61, opened 2018-09-26
`kill` nukes the docker registry


vmarkovtsev:
I run `srcd kill` and this is what it does:

```
INFO[0000] removing containers...                       
INFO[0000] removing container srcd-cli-gitbase          
INFO[0000] removing container srcd-cli-pilosa           
INFO[0001] removing container srcd-cli-bblfshd          
INFO[0002] removing container srcd-cli-daemon           
INFO[0003] removing volumes...                          
INFO[0003] removing volume srcd-cli-bblfsh-storage      
INFO[0006] removing images...                           
INFO[0006] removing image srcd/gitbase:latest           
INFO[0006] removing image bblfsh/bblfshd:latest         
INFO[0007] removing image srcd/style-analyzer:latest    
INFO[0007] removing image srcd/cli-daemon:latest        
INFO[0007] removing image srcd/gitbase-web:latest       
INFO[0007] removing image srcd/spark:2.2.1              
INFO[0007] removing image srcd/codelab:latest           
INFO[0009] removing image pilosa/pilosa:v0.9.0          
INFO[0009] removing image srcd/spark:2.2.0_v2
```

Removing containers is fine, but removing images o_O It erases bblfshd which I use frequently, spark (btw: two versions??) and style-analyzer which is innocent.

My suggestion is to not remove the images by default - it is very cruel.

marnovo:
Maybe to remove images something along the lines of the `prune` command you see in things like Homebrew?

campoy:
Yes, this should be fixed.

Who's the current maintainer of this, @mcuadros ?

vmarkovtsev:
Máximo wrote me in DM that Applications are responsible.

bzz:
> Hey @bzz, it seems you're responsible for this issue.

thanks for heads up @campoy and a quick fix @smacker !

I belive both are true since the beginning of the 2018Q4: 
 - App team maintains engine
 - I'm on LA team :)

smacker:
which request? the issue is fixed already

Issue #69, opened 2018-10-09
Add a softer `kill` that does not delete images


carlosms:
I may want to stop working on this right now, but I don't want to wait for the download process next time I continue. Maybe a `scrd stop`?

smacker:
fixed by making kill "softer" by default and introducing new flag to remove images too

PR #78, opened 2018-10-22
Kill: don't remove images by default

smacker:
Fix: #61 #69

As described in the issue the current behavior is very strange.
This PR changes the default. `srcd kill` won't remove images anymore by default. But only with new flag `--with-images`


carlosms:
This change requires to update the docs, at least in https://github.com/src-d/engine/blob/master/docs/commands.md.

smacker:
👍 updated.
grep shows it is the only place where `kill` command is mentioned

PR #88, opened 2018-10-24
Start component logging

### LINKED ISSUE #48: Logging installation of bblfsh drivers
The first time I run `srcd web sql` it will install the recommended drivers for bblfsh. I'd like the log to reflect which drivers are being downloaded & installed (as we do on `srcd init` with the cli-daemon).

Current:
```
srcd web sql --port=8088
INFO[0003] this is taking a while, if this is the first time you launch this web client, it might take a few more minutes while we install all the required images 
```

Desired:
```
srcd web sql --port=8088
INFO[0003] this is taking a while, if this is the first time you launch this web client, it might take a few more minutes while we install all the required images 
INFO[0004] installing "bblfsh/python-driver:latest"
INFO[0005] installed "bblfsh/python-driver:latest"
etc...
```

smacker:
Fix #48

This PR implements the common way to send logs from server to client and new method `StartComponentWithLogs` that uses this code.

cli command `srd web sql` calls it instead of `StartComponent`.

If we are ok with such output we can use `StartComponentWithLogs` in all cli commands.

Example of the output:
```
$ go run cmd/srcd/main.go web sql
INFO[0000] installing "srcd/cli-daemon:latest"
INFO[0005] installed "srcd/cli-daemon:latest"
INFO[0008] this is taking a while, if this is the first time you launch this web client, it might take a few more minutes while we install all the required images
INFO[0011] installing all recommended drivers
INFO[0013] installing bash driver version latest
INFO[0037] installing go driver version latest
INFO[0049] installing java driver version latest
INFO[0070] installing javascript driver version latest
INFO[0090] installing php driver version latest
INFO[0115] installing python driver version latest
INFO[0139] installing ruby driver version latest
```


campoy:
I was actually considering whether it would make sense to get rid of the `WithLogs` endpoints and instead use the docker API to fetch the logs from the container directly.

What do you think, @smacker and @smola?

smola:
Using Docker API for logs makes sense to me. In general, given this tool is so coupled to Docker, it does not make sense to reimplement functionality, possibly in an inferior way.

smacker:
I'm totally in for docker logs. Though I see a few problems with it:

1. We might want to have different logs on server and in cli.
Example of log for `sql` command on server:
```
time="2018-10-24T17:35:11Z" level=info msg="connecting to mysql \"root@tcp(srcd-cli-gitbase)/?maxAllowedPacket=65536\""
```
It doesn't appear on client and I don't really think it should.

2. Concurrent requests to the server
I can easily run 2 different srcd commands. Then logging will become a mess.

3. Different log levels on server and client

4. Lost of colors

If we are okay with all this issues we can move to getting logs from server container.

smacker:
Yes, there are workarounds for most of the issues. My point was just that reading logs from the container isn't as simple as it looks.

I'm closing this PR and create a new one based on it that would just pass context around.
As mentioned it's needed for concurrency and most probably will be useful in the future.

Issue #82, opened 2018-10-23
SQL: ResourceExhausted desc = grpc: received message larger than max


smacker:
```
$ scrd sql
gitbase> SELECT cf.file_path, f.blob_content
      -> FROM ref_commits r
      -> NATURAL JOIN commit_files cf
      -> NATURAL JOIN files f
      -> WHERE r.ref_name = 'HEAD'
      -> AND r.history_index = 0
      -> ;
rpc error: code = ResourceExhausted desc = grpc: received message larger than max (15214220 vs. 4194304)
```

Engine master, gitbase v0.17.1.

PR #90, opened 2018-10-24
Stream rows for SQL query

smacker:
Fix: #82


smacker:
This change breaks backward compatibility with old cli.
Before merging we must resolve https://github.com/src-d/engine/issues/92

smacker:
Cli-server compatibility issue was resolved.
It's also okay to break cli once: https://github.com/src-d/engine/issues/92#issuecomment-434800212

Merging.

PR #98, opened 2018-10-25
print detected message only once

campoy:
Currently the detected language will be printed once per tree returned, this can be many times when using a query, for instance "all identifiers"


campoy:
Done, @smacker

PR #99, opened 2018-10-25
avoid extra error message when leaving repl

campoy:
After `repl` we should simply exit the command.
Similarly empty statements should be ignored.


campoy:
Done, @smacker

PR #101, opened 2018-10-26
UAST2

smacker:
Use UAST2 for `parse` command.

`web parse` command from the last release already uses it
`gitbase` returns uast2 in the last beta release too


campoy:
Please try to keep your PRs independent in the future, I just finished reviewing #102 and there's a lot in common

smacker:
@campoy please read the description of PR before reviewing.

It's a common workflow for independent features (from a product point of view) that require a sequential change in code. We use it quite often in apps team. We have an agreement to describe the dependencies and what to review in the description of PR. In this case in #102 PR you can see:

> Based on #101 (last 2 commits are new)

such workflow:

- makes it easier to review changes
- avoids throw away work that would happen because of conflicts
- simplifies testing
- speedups merging

campoy:
I was not aware of that way of reviewing PRs, @smacker.

I'm still not sure how can I tell which changes of this PR were new or already included in other PRs.
Does that mean they're all new?

smacker:
@campoy we usually review like this:

If PR is based on another one go to base PR and review it first.
Let's use this 2 PRs as an example. You can see here 3 commits and the description why those changes were made. If everything looks good and you approve this PR, go back to the first PR.
There, in the PR description, you see only 2 new last commits are new. New means they don't exist a "base" PR. Everything before it you reviewed already. Then you go to "changed files" tab, in the top-left corner you select only "new" commits and review them.

Hope it makes sense for you. 👍

PR #109, opened 2018-10-30
Container logs

smacker:
fix: #48

output:
```
$ go run cmd/srcd/main.go web sql
INFO[0001] creating srcd-cli-network docker network
INFO[0005] this is taking a while, if this is the first time you launch this web client, it might take a few more minutes while we install all the required images
INFO[0005] starting bblfshd daemon
INFO[0005] installing all recommended drivers
INFO[0005] connecting to bblfsh management on srcd-cli-bblfshd:9433
INFO[0005] installing bash driver version latest
INFO[0039] installing go driver version latest
INFO[0057] installing java driver version latest
INFO[0083] installing javascript driver version latest
INFO[0109] installing php driver version latest
INFO[0140] installing python driver version latest
INFO[0168] installing ruby driver version latest
INFO[0193] starting gitbase web
```

product aka @campoy said in previous PR that we don't care about time or debug level, so I just get message from the logs and re-log it on cli side.


cmd/srcd/cmd/root.go
```diff
@@ -80,3 +85,35 @@ func initConfig() {
 		fmt.Println("Using config file:", viper.ConfigFileUsed())
 	}
 }
+
+var logMsgRegex, _ = regexp.Compile(`.*msg="(.+)"`)
+
+func logAfterTimeout(header string) chan struct{} {
+	logs, err := daemon.GetLogs()
+	if err != nil {
+		logrus.Fatalf("could get logs from server container: %v", err)
```
carlosms:
```suggestion
		logrus.Fatalf("could not get logs from server container: %v", err)
```

campoy:
I would log this only when `-v` is passed, otherwise it's gonna be a bit too noisy.

What do you think?

smacker:
It logs only if the process took more than 3 seconds. In case everything is cached/fast there will be no logs/noise.

I disagree we should log only on `-v` because my main pain point with lack of this feature is:
```
$ go run cmd/srcd/main.go web sql
this is taking a while, if this is the first time you launch this web client, it might take a few more minutes while we install all the required images
```
and waiting 10 minutes without any feedback what is going on. Does it work? Does it hang?
What's going on?????

campoy:
So will the logs appear if the query takes less than 10 seconds?

I'd say that if a query is taking more than that we should probably log something saying that it is probably because there's something being installed.
We can then indicate that next time you could run it with `-v` for more logs, or even interactively allow the user to enable the logs.

smacker:
I don't understand you. Which query? Why 10 seconds? As I said we start logging after 3 seconds (hard-coded value from before).
We still notify a user with user-friendly message that something is installing, you can see an example in PR description.

This PR doesn't implement interactivity or `-v`. It only solves #48 exactly as it described in the issue.

cmd/srcd/cmd/root.go
```diff
@@ -80,3 +85,35 @@ func initConfig() {
 		fmt.Println("Using config file:", viper.ConfigFileUsed())
 	}
 }
+
+var logMsgRegex, _ = regexp.Compile(`.*msg="(.+)"`)
```
campoy: Use `regexp.MustCompile(` instead of ignoring the error silently

cmd/srcd/cmd/root.go
```diff
@@ -80,3 +85,35 @@ func initConfig() {
...
+
+func logAfterTimeout(header string) chan struct{} {
+	logs, err := daemon.GetLogs()
+	if err != nil {
+		logrus.Fatalf("could not get logs from server container: %v", err)
```
campoy: Not sure whether this should be a `Fatalf`, maybe just `Errorf`?

cmd/srcd/cmd/root.go
```diff
@@ -80,3 +85,35 @@ func initConfig() {
...
+			for scanner.Scan() {
+				match := logMsgRegex.FindStringSubmatch(scanner.Text())
+				if len(match) == 2 {
+					logrus.Info(match[1])
+				}
+			}
+			if err := scanner.Err(); err != nil && err != context.Canceled {
+				logrus.Fatal(err)
+			}
+			logs.Close()
```
campoy: seems like you could move the` logs.Close` into a `defer logs.close` at the beginning of the goroutine

PR #130, opened 2018-11-14
add help and validation for positional arguments

### LINKED ISSUE #126: CLI help does not show positional arguments
Several commands like `sql` or `parse drivers install` can use or require a parameters but is not shown. Also when it is required (`parse drivers install`) does not show error:

```
$ srcd sql --help
Run a SQL query over the analyzed repositories.

Usage:
  srcd sql [flags]

Flags:
  -h, --help   help for sql

Global Flags:

$ srcd parse drivers install --help
Install language drivers.

Usage:
  srcd parse drivers install [flags]

Flags:
  -h, --help   help for install

Global Flags:

$ srcd parse drivers install ; echo $?
0
```

smacker:
Fix #126

### DIFF
--- a/cmd/srcd/cmd/components.go
+++ b/cmd/srcd/cmd/components.go
@@ -49,8 +49,9 @@ var componentsListCmd = &cobra.Command{
 
 // componentsCmd represents the components install command
 var componentsInstallCmd = &cobra.Command{
-	Use:   "install",
+	Use:   "install [component]",
 	Short: "Install source{d} component",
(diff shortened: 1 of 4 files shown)


carlosms:
I wonder if we should at some point move to https://github.com/src-d/go-cli, for consistency with other projects.

smacker:
`go-cli` uses `go-flags` and `engine` uses `cobra` 🤷‍♂️

carlosms:
yes, that's why I wonder if we should migrate, for consistency. Not a priority, of course.

PR #134, opened 2018-11-20
[WIP][DISCARDED] Fix container dependencies for init cmd

### LINKED ISSUE #60: `init` does not re-initialize
If I run `srcd init` on new directory, it does not restart the daemon container, so the new path is not applied.

carlosms:
Fix #60.

The code in  `Kill()` was there to stop the working directory dependencies, but somehow the image names and container names were mixed.
`Kill` needs container names for `RemoveContainer`, and `List` was returning image names. Besides that `KnownComponents` could not work because it was comparing image names to container names.

### DIFF
--- a/cmd/srcd/cmd/components.go
+++ b/cmd/srcd/cmd/components.go
@@ -42,7 +42,7 @@ var componentsListCmd = &cobra.Command{
 			return err
 		}
 
-		imgs, err := components.List(
+		imgs, err := components.ImageList(
 			context.Background(),
 			components.KnownComponents(daemonVersion, allVersions))
 		if err != nil {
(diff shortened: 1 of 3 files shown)


carlosms:
I see what you mean. But take a look here: https://github.com/src-d/engine/pull/134/files#diff-f65e86a43ab65d559e1ce8134b4aa922R58

The output of the previous `List` was being used as input for `docker.RemoveContainer`, so it's important to know what you are getting. Previously it was image names, so it was impossible that would work.

I would agree with you if the output of `components.List` was `[]Component`, and then you had `func (c *Component) String()` and `func (c *Component) Stop()`. And maybe we can refactor things to work that way, but I don't know if we want to do that in this PR, or later in a different PR, or put that on hold until the long-term roadmap is more clear...

smacker:
In this PR you change the API of `components` module a lot.

I would prefer to change it with `components.List` returning `[]Component` than current changes that break abstraction.

The amount of work should be the same (just instead of renaming `List` to `ListImages` you will need to pass `component.Image`)

smacker:
One more argument in support of `components.List` instead of `ListImages/ListContainers`:
No need for 2 different filter function types and 2 implementations of `KnownComponents`.

carlosms:
This PR has been replaced by #138 & #139. I'll close this one if the others are merged.

Issue #67, opened 2018-10-09
Use `latest-drivers` bblfsh image


carlosms:
With the current [bblfsh v2.90](https://github.com/bblfsh/bblfshd/releases/tag/v2.9.0) a `bblfshctl driver install` fails for already installed drivers (context [here](https://github.com/bblfsh/bblfshd/issues/195)). 

This means we can get errors like this:
```
$ srcd parse uast ~/repos/gitbase-web/cmd/gitbase-web/main.go 
FATA[0000] could not stream: rpc error: code = AlreadyExists desc = driver already installed: go (image reference: docker://bblfsh/go-driver:latest): %s 
```

We should add the  new `--force` flag to `bblfshctl` (which actually means ignore, not force), or use the new `latest-drivers` images.

smacker:
The error mentioned in the issue is fixed here: https://github.com/src-d/engine/pull/84
But we still can consider using `latest-drivers` image.

smacker:
I have noticed one problem with `latest-drivers`. This image follows bblfshd release cycle. So it doesn't get updates when bugs are fixed in drivers or even new features are added.

As an example right now the latest drivers image is `v2.9.1-drivers` https://hub.docker.com/r/bblfsh/bblfshd/tags/ released 15 days ago.
But the java-driver is released 20 hours ago https://hub.docker.com/r/bblfsh/java-driver/tags/ 
It is 2 releases newer with changes:
- Fix roles for primitive types
- Support error codes

smacker:
In my opinion, we need to discuss it with LA team and agree on some kind of schedule for `-drivers` releases or something like that.

Current release was made 24 days ago, since then in java-driver, for example, was created 3 new releases:
- Enable language namespace in Semantic UAST
- Annotate default function modifier.
- Normalize and annotate Javadoc nodes properly.

We need to incorporate those improvements into engine somehow at some point.

cc @creachadair

creachadair:
The above sounds sensible to me. At the very least, it seems like we should have some coordination around which semantic version ranges are compatible with each other. Tightly-coupled releases are hard to maintain, but there should be some middle ground between that and confusing/surprising behaviour for the user.

smacker:
here is the change to `-drivers` image: https://github.com/src-d/engine/pull/135

but it still downloads cpp driver separately. Because cpp driver is recommended already but isn't included in latest `drivers` image. (one more example why we need some kind of coordination with LA team)

PR #135, opened 2018-11-20
use bblfsh-drivers image

smacker:
Fix: #67

### DIFF
--- a/components/components.go
+++ b/components/components.go
@@ -48,7 +48,7 @@ var (
 	Bblfshd = Component{
 		Name:    "srcd-cli-bblfshd",
 		Image:   "bblfsh/bblfshd",
-		Version: "v2.9.1",
+		Version: "v2.9.2-drivers",
 	}
 
 	BblfshWeb = Component{


carlosms:
I would image that with `-drivers` image we would also remove the drivers installation.

As it is now, if we are using the bigger image, and still re-install all drivers, does it make sense? Are we not worsening the current situation?

smacker:
I'm not sure I understand your concern. I did quick test and it worked as expected. (slowly download bblfsh-drivers image and then install all drivers except cpp super quick)

Could you please provide steps how to reproduce the problem you are talking about.

carlosms:
It was more a question than a bug report. I've done some testing and now I see that even if the driver install call is made for `latest`, the installed drivers are not updated to the latest release, they just return "driver already installed".

carlosms:
What about this comment you made https://github.com/src-d/engine/issues/67#issuecomment-433346503?

> Right now we have very good example of why it's problem:
> We have bblfsh-web v0.8.0 but it can't work with current latest drivers image. Only with a normal one because it requires newer drivers.

If the drivers that ship bundled with bblfsh v2.9.2-drivers are not updated, will this PR break https://github.com/src-d/engine/pull/136?

smacker:
> It was more a question than a bug report. I've done some testing and now I see that even if the driver install call is made for latest, the installed drivers are not updated to the latest release, they just return "driver already installed".

Now I get what you mean. Yes, bblfshd works in the same way as docker. If it has `latest` tag locally it just uses it without checking for a newer one on the remote.

> What about this comment you made #67 (comment)?

it was about version v2.9.0 or v2.8.0, I don't remember which one was the latest for the moment of me writing that comment. v2.9.2 works correctly with bblfsh-web.

carlosms:
great, thanks for confirming it will work fine

PR #141, opened 2018-11-23
Fix version listing for images with more than one tag

carlosms:
Bug introduced in PR #138.


carlosms:
I agree, I guess once the long-term roadmap is there we'll take care of testing.

PR #142, opened 2018-11-23
Use Components for srcd components install

carlosms:
Using `Component`s for `srcd components install` means that we allow to install only the exact images (and versions) engine manages, instead of any image with the `srcd/` or `bblfsh/` prefix as we do in master.

With this PR we also get a nicer error when the component to install is not found:
```
$ srcd install srcd/lookout
Error: srcd/lookout is not valid. Component must be one of [srcd/cli-daemon, srcd/gitbase, srcd/gitbase-web, bblfsh/bblfshd, bblfsh/web]
```

### DIFF
--- a/cmd/srcd/cmd/components.go
+++ b/cmd/srcd/cmd/components.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/spf13/cobra"
@@ -51,7 +52,7 @@ var componentsListCmd = &cobra.Command{
 		for _, cmp := range cmps {
 			fmt.Fprintf(w, "%s\t%s\t%v\t%v\n",
 				cmp.ImageWithVersion(),
-				boolFmt(cmp.IsInstalled(context.Background())),
+				boolFmt(cmp.IsInstalled()),
 				boolFmt(cmp.IsRunning()),
 				cmp.Name,
 			)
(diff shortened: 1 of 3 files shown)


cmd/srcd/cmd/components.go
```diff
@@ -60,30 +60,56 @@ var componentsInstallCmd = &cobra.Command{
...
 
-			if ok {
-				log.Printf("%q is already installed", arg)
+			if c == nil {
+				names := make([]string, len(cmps))
+				for i, cmp := range cmps {
+					names[i] = cmp.Image
+				}
+
+				return fmt.Errorf("%s is not valid. Component must be one of [%s]", arg, strings.Join(names, ", "))
```
campoy: This might create a huge error message, maybe it's worth saying it's not an authorized image and how to find the list of valid ones?

components/components.go
```diff
@@ -159,18 +160,15 @@ var IsWorkingDirDependant FilterFunc = func(cmp Component) (bool, error) {
...
-	return docker.IsInstalled(ctx, cmp.Image, cmp.Version)
+var IsInstalled FilterFunc = func(cmp Component) (bool, error) {
+	return cmp.IsInstalled()
 }
 
-// IsRunningFilter filters Components that have a container running, using
+// IsRunning filters Components that have a container running, using
 // its image with the exact version
-var IsRunningFilter FilterFunc = func(cmp Component) (bool, error) {
+var IsRunning FilterFunc = func(cmp Component) (bool, error) {
```
campoy: Why are these variables instead of standard functions?

cmd/srcd/cmd/components.go
```diff
@@ -60,30 +60,56 @@ var componentsInstallCmd = &cobra.Command{
...
 
-			log.Printf("installing %s", arg)
+			log.Printf("installing %s", c.ImageWithVersion())
 
-			if err := components.Install(context.Background(), arg); err != nil {
-				log.Printf("could not install %s: %v", arg, err)
-				os.Exit(1)
+			err = c.Install()
+			if err != nil {
+				log.Printf("could not install %s: %s", arg, err)
```
campoy: shouldn't this return an error?

cmd/srcd/cmd/components.go
```diff
@@ -60,30 +60,56 @@ var componentsInstallCmd = &cobra.Command{
...
+			if c == nil {
+				names := make([]string, len(cmps))
+				for i, cmp := range cmps {
+					names[i] = cmp.Image
+				}
+
+				return fmt.Errorf("%s is not valid. Component must be one of [%s]", arg, strings.Join(names, ", "))
+			}
+
+			_, err = c.RetrieveVersion()
```
smacker: this function would update version of the component right?
So if I do `srd components install srcd/some:0.1.0` it will actually install `srcd/some:0.1.298`?
It's unexpected.

cmd/srcd/cmd/components.go
carlosms: I don't think it's that big, compared to the huge help that follows below it, and it saves an extra step for the user to find that list:
```
$ ./srcd components install nope
Error: nope is not valid. Component must be one of [srcd/cli-daemon, srcd/gitbase, srcd/gitbase-web, bblfsh/bblfshd, bblfsh/web]
Usage:
  srcd components install [component] [flags]

Flags:
  -h, --help   help for install

Global Flags:

nope is not valid. Component must be one of [srcd/cli-daemon, srcd/gitbase, srcd/gitbase-web, bblfsh/bblfshd, bblfsh/web]
```

Where would you propose to put the list? The online docs? The `-h` output?

components/components.go
carlosms: To ensure that they are `FilterFunc` type, and document they are intended to be used as `FilterFunc`. Is it not idiomatic?

cmd/srcd/cmd/components.go
carlosms: thanks, fixed in dcf9c2f.
I'll squash before merging.

cmd/srcd/cmd/components.go
carlosms: This is here to mimic the behaviour of `srcd init`. It's true that it will be confusing if the user includes the version in the argument, so I removed it in d85bdd7.
Now you won't be able to install `some:0.1.0`, it will not be a valid name.

cmd/srcd/cmd/components.go
campoy: Actually, the error is not that bad, indeed.

I'm not sure we should be returning the error since it causes the whole `usage` message to be printed too.

components/components.go
campoy: not really, no

the more idiomatic way would be to have:

```go
func IsRunning() FilterFunc  {
   return func(cmp Component) (bool, error) { ... }
}
```

cmd/srcd/cmd/components.go
carlosms: Yes, I also dislike the whole `--help` output for a wrong parameter, but it's consistent with other `src-d` commands that use https://github.com/src-d/go-cli.

components/components.go
carlosms: > the more idiomatic way would be to have:
> 
> ```go
> //  IsRunning is a FilterFunc...
> func IsRunning(cmp Component) (bool, error)  {
>    ...
> }
> ```

I've included that in the squashed changes before merging

PR #145, opened 2018-11-27
docs: Remove reference to "srcd parse native".

creachadair:
This command is not implemented; instead (it appears) the intended use is to
pass "-m native" to the "srcd parse uast" subcommand.

Signed-off-by: M. J. Fromberger <michael.j.fromberger@gmail.com>


smacker:
thanks!

PR #147, opened 2018-11-30
Make daemon version update failure a warning instead of error

### LINKED ISSUE #146: Commands fail if the engine version information cannot be retrieved
For example, if there is no internet connectivity, this happens:
```
$ ./srcd sql "select * from repositories"
2018/11/28 11:45:28 could not get daemon client: unable to get compatible daemon version: can't authorize in docker registry: Get https://auth.docker.io/token?scope=repository%3Asrcd%2Fcli-daemon%3Apull&service=registry.docker.io: dial tcp: lookup auth.docker.io on 127.0.0.53:53: server misbehaving
```

carlosms:
Fix #146.


cmd/srcd/daemon/daemon.go
```diff
@@ -110,9 +110,10 @@ func start(workdir string) (*docker.Container, error) {
 		return nil, errors.Wrap(err, "unable to get home dir")
 	}
 
-	tag, hasNew, err := docker.GetCompatibleTag(components.Daemon.Image, cliVersion)
+	cmp := components.Daemon
+	hasNew, err := cmp.RetrieveVersion()
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to get compatible daemon version")
+		logrus.Warn("unable to get compatible daemon version from Docker Hub: ", err)
```
smacker: I wrote this error message but now I see it's kind of misleading. When I read it, it looks like it can't get the daemon itself. Not just check for new version. Can we change it something like "unable to receive information about engine update from Docker Hub"? Or maybe you can find better wording for it.
Thanks!

carlosms: True, it can be read as both getting the version info, or the new image. What about
"failed to list the available daemon versions on Docker Hub: "

carlosms: done, change rebased and force-pushed

PR #184, opened 2018-12-26
Just some improvements on init.go

se7entyse7en:
By using the standard `Args` key of Cobra `Command`, the error message in case of a number of arguments > 1 changes from:

```
✘  se7entyse7en in ~/Projects/src-d/engine/cmd/srcd (init-cmd-clean)  $ srcd init foo bar                                                                                        [12/26/18 | 13:06:40.051]
FATA[0000] invalid number of arguments given, expecting 0 or 1
```

to:

```
✔  se7entyse7en in ~/Projects/src-d/engine/cmd/srcd (init-cmd-clean)  $ go run main.go init foo bar                                                                              [12/26/18 | 13:06:17.418]
Error: accepts at most 1 arg(s), received 2
Usage:
  srcd init [flags]

Flags:
  -h, --help   help for init

Global Flags:

accepts at most 1 arg(s), received 2
exit status 1
```

### DIFF
--- a/cmd/srcd/cmd/init.go
+++ b/cmd/srcd/cmd/init.go
@@ -28,11 +28,8 @@ import (
 var initCmd = &cobra.Command{
 	Use:   "init",
 	Short: "Starts the daemon or restarts it if already running.",
+	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		if len(args) > 1 {
-			logrus.Fatal("invalid number of arguments given, expecting 0 or 1")
-		}
-
 		err := daemon.Kill()
 		if err != nil {
 			logrus.Fatal(err)
--- a/cmd/srcd/cmd/init.go
+++ b/cmd/srcd/cmd/init.go
@@ -30,12 +30,9 @@ var initCmd = &cobra.Command{
 	Short: "Starts the daemon or restarts it if already running.",
 	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		err := daemon.Kill()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/srcd/cmd/init.go
```diff
@@ -28,39 +28,35 @@ import (
...
+		if err != nil {
+			logrus.Fatal(err)
 		}
 
 		info, err := os.Stat(workdir)
 		if err != nil || !info.IsDir() {
 			logrus.Fatalf("path %q is not a valid working directory", workdir)
 		}
 
+		err = daemon.Kill()
```
se7entyse7en: I think that it's better to kill the daemon only after all the validation has succeeded.

cmd/srcd/cmd/init.go
```diff
@@ -28,39 +28,35 @@ import (
...
-			}
+		}
+		if err != nil {
+			logrus.Fatal(err)
 		}
 
 		info, err := os.Stat(workdir)
 		if err != nil || !info.IsDir() {
 			logrus.Fatalf("path %q is not a valid working directory", workdir)
 		}
```
se7entyse7en: Actually I think that everything until this line should go inside a custom function provided to the `Args` key given that it still contains validation logic. But it would require recomputing the `workdir` variable. One way to avoid this is to define `var workdir string` at top-level, and it will be initialized during the args validation.

carlosms: What you mention might be cleaner if we had to reuse the same validation. But the current code is small, maybe it's not worth it to have a package variable just for that.

se7entyse7en: yup exactly :)

se7entyse7en:
🎉

Issue #85, opened 2018-10-23
Remove driver management commands


smacker:
From the docs:

> All of the subcomands of srcd parse drivers provide management for the language drivers installed on bblfsh.

but it's marked as TBD.

We need to test it and update docs if everything works.

smacker:
I agree with you in general but I think ability to install experimental or 3rd-party drivers may be handy too. 

@campoy

campoy:
I think that not having the ability to install experimental stuff on the engine is totally fine.
The whole idea behind the engine is to provide an easy way to get started for anyone.

We will discuss this and more on a meeting tomorrow, I added you (@carlosms) to the meeting.
It'd be great if you can make it, I think your opinion will help us better decide what can be done.

smacker:
I updated the issue title. It makes much more sense now as we don't install drivers directly anymore but use `bblfsh-drivers` image.

PR #188, opened 2019-01-08
Remove driver commands

smacker:
Fix: #85 

I removed it from both cli&server but in separate commits.

I don't think we should have dead code in the repo. In case we will need for some reason to manage drivers again on server we can just revert the commit.


se7entyse7en:
> I don't think we should have dead code in the repo. In case we will need for some reason to manage drivers again on server we can just revert the commit.

agree

Issue #152, opened 2018-12-04
srcd sql installs dependencies only after the first query is sent


smacker:
```
$ ./srcd sql
gitbase> show tables;
INFO[0012] installing "srcd/cli-daemon:v0.4.0"
INFO[0018] installed "srcd/cli-daemon:v0.4.0"
```

In my case, it had a lot of time to install everything before I typed the first query because I was reading docs. But current behavior caused unnecessary waiting.

PR #189, opened 2019-01-08
Sql install deps

smacker:
Fix: #152

1st commit implements installation before starting repl.
2nd commit fixes a bug that we still read logs after started channel is closed.

Example output:
```
$ go run cmd/srcd/main.go sql
INFO[0005] this is taking a while, if this is the first time you launch sql client, it might take a few more minutes while we install all the required images
INFO[0005] installing \"bblfsh/bblfshd:v2.11.0-drivers\"
INFO[0102] installed \"bblfsh/bblfshd:v2.11.0-drivers\"
INFO[0102] starting bblfshd daemon
gitbase> select * from repositories;
+---------------+
| REPOSITORY ID |
+---------------+
| repos         |
+---------------+
```

### DIFF
--- a/cmd/srcd/cmd/sql.go
+++ b/cmd/srcd/cmd/sql.go
@@ -52,7 +52,9 @@ var sqlCmd = &cobra.Command{
 			logrus.Fatalf("could not get daemon client: %v", err)
 		}
 
-		started := logAfterTimeout("this is taking a while, if this is the first time you launch sql client, it might take a few more minutes while we install all the required images")
+		started := logAfterTimeout("this is taking a while, " +
+			"if this is the first time you launch sql client, " +
+			"it might take a few more minutes while we install all the required images")
 
 		// Might have to pull some images
 		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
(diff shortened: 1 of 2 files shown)


smacker:
CI is actually green if you click on "details". The status wasn't updated on GH.

smacker:
ping @dpordomingo @se7entyse7en

cmd/srcd/cmd/sql.go
```diff
@@ -45,6 +47,24 @@ var sqlCmd = &cobra.Command{
 			query = args[0]
 		}
 
+		c, err := daemon.Client()
+		if err != nil {
+			logrus.Fatalf("could not get daemon client: %v", err)
+		}
+
+		started := logAfterTimeout("this is taking a while, if this is the first time you launch sql client, it might take a few more minutes while we install all the required images")
```
dpordomingo: I'd define a constant with this text, and use it here; or at least split the line to fit in the line limit.

smacker: splitted. Defining a constant that is used only once and is warning message doesn't sound good for me at all...

dpordomingo: good enough. Now it's readable in GH :dagger: thanks

PR #210, opened 2019-02-11
Add checks for generated code in Travis

carlosms:
Similar to the checks we do in lookout.
For go modules and proto files.


.travis.yml
```diff
@@ -12,8 +12,18 @@ matrix:
...
+  include:
+    - name: 'Unit Tests'
+      script:
+        - make test-coverage
+    - name: 'Generated code'
+      script:
+        - GO111MODULE=on go mod tidy
+        - GO111MODULE=on go mod vendor
+        - make no-changes-in-commit
+        - (cd ./api && make)
```
dpordomingo: I'd make it explicit:
```suggestion
        - make -C api proto
```

`make` without target is tricky, and can vary without noticing.

carlosms: Thanks, change applied and squashed

PR #234, opened 2019-02-26
Adds support for windows by fixing host path for mounted volumes

### LINKED ISSUE #45: Windows support
Currently on Windows, when running `srcd init C:\path\etc` the engine hangs on installing the cli daemon.

se7entyse7en:
Closes #45.

Adds support for windows.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


docker/docker.go
```diff
@@ -264,6 +265,31 @@ func WithSharedDirectory(hostPath, containerPath string) ConfigOption {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	info, err := c.Info(ctx)
+	if err != nil {
+		logrus.Fatalf("could not get information")
+	}
+
+	isWinHost := info.OSType == "windows" || strings.Contains(
```
se7entyse7en: Initially, I thought that the first check was enough. But for example in the Windows Server we're using the "OSType" is actually equal to "linux" and the "OperatingSystem" is "Docker for Windows". I think that it is because it's actually the machine being run by Hyper-V, but I'm guessing. I didn't find any specific documentation about this case.

docker/docker.go
```diff
@@ -264,6 +265,31 @@ func WithSharedDirectory(hostPath, containerPath string) ConfigOption {
...
+		strings.ToLower(info.OperatingSystem), "windows")
+	if isWinHost {
+		// For Windows we need to change paths as
```
se7entyse7en: This is actually something that I found when I worked on this issue the first time, but I cannot find the issue anymore in the docker for win repo. 😞

se7entyse7en:
This still lacks of README.md update and documentation. I've opened this to have your opinions about the solution.

Also, I tried it by building the docker image of the daemon directly on Windows, and by running engine directly with `go run .\cmd\srcd\main.go sql "select 1"`. I'm gonna try using the exe, but I don't think that is goind to misbehave (I hope).

docker/docker.go
```diff
@@ -264,6 +265,31 @@ func WithSharedDirectory(hostPath, containerPath string) ConfigOption {
...
+		// C:\\Users\\Windows10\\go\\src\\github.com\\src-d\\engine to
+		// //c/Users/Windows10/go/src/github.com/src-d/engine
+		hostPath = strings.Replace(hostPath, "\\", "/", -1)
+		hostPath = regexp.MustCompile(`^(\w):`).ReplaceAllStringFunc(hostPath, func(m string) string {
```
dpordomingo: Is there no way to rely on go [`path/filepath`](https://golang.org/pkg/path/filepath/) to generate proper paths?

se7entyse7en:
I'm gonna close this for now as I maybe found a better solution 👍.

se7entyse7en:
Closed in favor of #235.

docker/docker.go
se7entyse7en: Please take a look at #235. BTW I wasn't able to find a better way to do this.

PR #235, opened 2019-02-27
Adds Windows support

### LINKED ISSUE #45: Windows support

se7entyse7en:
Closes #45.

Adds Windows support. This was tested on Windows 10 Pro.


README.md
```diff
@@ -63,6 +63,8 @@ Follow these instructions:
   sudo pacman -S docker
   ```
```
dpordomingo:
```suggestion
```

README.md
```diff
@@ -63,6 +63,8 @@ Follow these instructions:
...
 
+- [Docker for Windows](https://hub.docker.com/editions/community/docker-ce-desktop-windows). Make sure to read the system requirements [here](https://docs.docker.com/docker-for-windows/install/).
+
 #### 2. Install source{d} Engine
 
 Download the **[latest release](https://github.com/src-d/engine/releases)** for MacOS (Darwin) or Linux.
```
dpordomingo:
```suggestion
Download the **[latest release](https://github.com/src-d/engine/releases)** for MacOS (Darwin), Linux or Windows.
```

README.md
```diff
@@ -91,25 +93,45 @@ sudo mv engine_linux_amd64/srcd /usr/local/bin/
 
 **Windows:**
 
-_Windows support is coming soon!_
+Extract the tar file with the tool you prefer.
+
+Open your powershell and move the `srcd.exe` file it into any path listed in `PATH`.
```
dpordomingo:
```suggestion
Copy  `srcd.exe` into a directory pointed by your `%path%` environment variable to make it available from anywhere.
```

README.md
```diff
@@ -91,25 +93,45 @@ sudo mv engine_linux_amd64/srcd /usr/local/bin/
...
+Move it into your local bin folder to be executable from anywhere:
+```powershell
+mv engine_windows_amd64\srcd.exe some\path
+```
+
+You can list the paths by running the following command:
```
dpordomingo:
```suggestion
You can list the directories pointed by your `%path%` environment variable by running the following command:
```

README.md
```diff
@@ -91,25 +93,45 @@ sudo mv engine_linux_amd64/srcd /usr/local/bin/
 
 **Windows:**
 
-_Windows support is coming soon!_
+Extract the tar file with the tool you prefer.
+
+Open your powershell and move the `srcd.exe` file it into any path listed in `PATH`.
+Move it into your local bin folder to be executable from anywhere:
```
dpordomingo: dupe
```suggestion
```

README.md
carlosms: Can we simplify this section? Is there a default path we can use here to make it ready to copy&paste?

docker/docker.go
```diff
@@ -264,6 +265,30 @@ func WithSharedDirectory(hostPath, containerPath string) ConfigOption {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	info, err := c.Info(ctx)
+	if err != nil {
+		logrus.Fatalf("could not get information")
+	}
+
+	isWinHost := info.OSType == "windows" || strings.Contains(
```
dpordomingo: I'd extract this functionality  in a separate function, like
```go
hostPath = mountSource(info, hostPath)
```
that would return the proper mount source for the passed OS

docker/docker.go
```diff
@@ -264,6 +265,30 @@ func WithSharedDirectory(hostPath, containerPath string) ConfigOption {
...
+	if err != nil {
+		logrus.Fatalf("could not create docker client")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	info, err := c.Info(ctx)
+	if err != nil {
+		logrus.Fatalf("could not get information")
```
carlosms: fatal feels a bit too drastic, this will kill the daemon. If we can I think it would be nice to return a nice error with the reason why the container could not be created.

README.md
se7entyse7en: Here's the content of `PATH` on the dev machine:
```powershell
PS C:\Users\Windows10\go> $env:path.split(";")
C:\ProgramData\DockerDesktop\version-bin
C:\Program Files\Docker\Docker\Resources\bin
C:\Windows\system32
C:\Windows
C:\Windows\System32\Wbem
C:\Windows\System32\WindowsPowerShell\v1.0\
C:\Windows\System32\OpenSSH\
C:\Program Files\Git\cmd
C:\Go\bin
C:\ProgramData\chocolatey\bin
C:\Users\Windows10\AppData\Local\Microsoft\WindowsApps
C:\Users\Windows10\go\bin
```

I think that `C:\Windows\system32` could be a nice candidate.

dpordomingo: When I was young, most of apps suggested you add your directory into your path env... It could be an alternative...
```powershell
SET PATH=%PATH%;{DIR_WHERE_SRCD.EXE_WAS_UNPACKED};
```

se7entyse7en: @dpordomingo yup that is an alternative. I avoided explaining that just to make the quickstart easier.

dpordomingo: manually copying a downloaded thing into `system32`? wow... it might be scary...

docker/docker.go
se7entyse7en: you're right, it seems a bit drastic. anyway, I then need to change the return value to `(ConfigOption, error)`. But what about the other functions returning a `ConfigOption`? Maybe for consistency, we should change them too even if they don't raise any potential error.

README.md
se7entyse7en: > it might be scary...
😂 
I actually don't know windows, but I guess it doesn't differ too much from copying a downloaded thing into `/usr/local/bin` in unix systems. But maybe I'm wrong.

dpordomingo: [Go installation guide](https://golang.org/doc/install/source#install) in Windows suggests you to add go bin directory in your `PATH`, so it seems a default behavior for windows developers...

dpordomingo: I think that `C:\Windows\system32` is not a user directory. I wonder if even the antivirus could complain about it...

se7entyse7en: Ok. We just have to decide where to put it then.

se7entyse7en: > I wonder if even the antivirus could complain about it...

Good point! 👍 
I haven't thought about that, it's been years that I don't run an antivirus...

se7entyse7en: What about `C:\Program Files\Engine` (maybe too generic)? or `C:\Program Files\srcd`?

dpordomingo: If there is no permissions problem, both lgtm

carlosms: I prefer `C:\Program Files\srcd`, it sounds more generic, and it could be used for other tools we may release in the future.

docker/docker.go
carlosms: :thinking: yes for consistency, but then everything gets flooded with `if err != nil`... I think for now we can modify the return values only for those that can return errors.

Or, if returning an error here makes the code uglier, you could also make the path translation as a separate step. Something like this:
```go
	indexDir := filepath.Join(s.datadir, "gitbase", s.workdirHash)

	workdirHostPath, err := docker.HostPath(s.workdir)
	if err != nil {
	}

	indexHostPath, err := docker.HostPath(indexDir)
	if err != nil {
	}

	return Component{
		Name: gitbase.Name,
		Start: createGitbase(
			docker.WithSharedDirectory(workdirHostPath, gitbaseMountPath),
			docker.WithSharedDirectory(indexHostPath, gitbaseIndexMountPath),
			docker.WithPort(gitbasePort, gitbasePort),
		),
		Dependencies: []Component{
			s.bblfshComponent(),
		},
	}
```

se7entyse7en: Addressed [here](https://github.com/src-d/engine/pull/235/commits/0db388d158478a338ef9c371f79e278d6633876c). The code has now more `if err != nil`, but I think is better as-is as it surely provides a more informative error.

PR #236, opened 2019-02-27
Config file with containers ports

### LINKED ISSUE #233: Port management config file
Task to implement the [Engine Port Management design doc](https://docs.google.com/document/d/17_ArQ4n7J7C_j7r9h27_hfTTWtX4HAwkORAibiLZdKk/edit#).

carlosms:
Fix #233.

I moved the config file initialization to a new package to avoid an import loop.

The Config contents are in `api` because even if it's not part of the gRPC api, clients should still follow the required format. So I thought it made sense to include it in the same package.

You'll see the `srcd web` commands read the config file too. I did this because we need to know the port in the CLI, to open the browser. The other option I considered was to send the `StartComponentRequest` without setting the `Port`, and return the port used by the new container in `StartComponentResponse`. I went for the most straightforward option, we can change this later if needed.


carlosms:
I've pushed the change in of `StartComponentResponse` mentioned in the description, 8a790ee.

@dpordomingo about moving the port constants, I'd prefer to limit the scope of this PR and merge fast. Please open a new issue for it and we can discuss it there.

se7entyse7en:
@carlosms Thanks for already addressing the port thing! 👍

carlosms:
Rebased on top of master, let's merge

PR #243, opened 2019-02-28
Use travis stage for deployment, skip latest docker tag for pre-releases

carlosms:


carlosms:
Sorry, I didn't mean to merge. Tagging @src-d/applications for a quick look in case I missed something.

The goal of the PR is to:
- Make the deployment a separate stage.
  - The previous stage run `deploy` & `after_deploy` once for each job, see [one](https://travis-ci.com/src-d/engine/jobs/181308103#L294), [two](https://travis-ci.com/src-d/engine/jobs/181308104#L468).
  - If one of the test stage jobs fail, now the deployment stage will not be executed
- Do not tag pre-release docker images as `latest`. To do that we must use `docker-push-latest-release`, as [documented here](https://github.com/src-d/ci/tree/v1#docker-build-and-push)

smacker:
Looks great! Thank you.

Issue #244, opened 2019-02-28
Add --log-format flag


se7entyse7en:
Adding the `--logfmt` could simplify the parsing of the logs during integration testing by running the cli commands with `--logfmt json` that is then propagated to [logrus formatter](https://github.com/sirupsen/logrus#formatters).

/cc @src-d/product given that the cli is user-facing.

smacker:
I propose to change logging/cli to the standard library we use in src-d.
Then we will have `log-format` option: https://github.com/src-d/go-cli/blob/master/log.go#L11
And it will be consistent with all other projects.

carlosms:
Yup, I brought up the idea of switching from cobra to go-cli with @smola but he thought it was not worth the extra effort to migrate.
@smola can you comment on this?

smola:
You could start by migrating to `src-d/go-log` and adding the `--log-format` (not `--logfmt`) flag.

@carlosms  With respect to `src-d/go-cli`, last time I did this, migrating from a cobra-based CLI to a go-cli one took me a while, and it was much smaller than this one. Do you have any rough estimation of how much could this take?

carlosms:
@smola I did a quick and dirty test, migrating `srcd init`

I don't want to be super optimistic, but the migration was not painful at all. There might be a few things that won't match. For example, go-cli already has a version command, we have our own in Engine.

I'd say this is not  a huge task. A few _working_ hours, meaning a few _real_ days depending on what other tasks we have to deal with.

See
- https://github.com/carlosms/engine/tree/go-cli/cmd
- https://github.com/src-d/engine/compare/master...carlosms:go-cli

smola:
@carlosms Let's go for it then ;)

carlosms:
Done in #410.

PR #302, opened 2019-03-06
Adds integration testing

se7entyse7en:
Adds integration tests that can be run as follows:
```
make test-integration
```

To avoid interacting with eventually running containers, the names of the containers used for running integration testing have the `integration-testing-scrd-cli-` prefix.

A new image of the deamon is built before the integration testing with the name `srcd/cli-daemon:integration-testing`.

As first suite, this PR adds an integration test for the `srcd init` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,3 +14,25 @@ $(MAKEFILE):
 	git clone --quiet --branch $(CI_VERSION) --depth 1 $(CI_REPOSITORY) $(CI_PATH);
 
 -include $(MAKEFILE)
+
+GOTEST_INTEGRATION_TAGS_LIST = integration
+GOTEST_INTEGRATION_TAGS = $(GOTEST_INTEGRATION_TAGS_LIST)
+GOTEST_INTEGRATION = $(GOTEST) -parallel 1 -count 1 -tags="$(GOTEST_INTEGRATION_TAGS)"
+
+INTEGRATION_TEST_BUILD_PATH = "build-integration"
+INTEGRATION_TEST_BIN_PATH = $(INTEGRATION_TEST_BUILD_PATH)/bin
+
+GOBUILD_INTEGRATION = $(GOCMD) build -tags "$(GOTEST_INTEGRATION_TAGS)"
+
+clean-integration:
+	rm -rf $(INTEGRATION_TEST_BUILD_PATH)
+build-integration: BUILD_PATH=$(INTEGRATION_TEST_BUILD_PATH)
+build-integration: BIN_PATH=$(INTEGRATION_TEST_BIN_PATH)
+build-integration: GOBUILD=$(GOBUILD_INTEGRATION)
+build-integration: build
+
+build-integration-daemon:
+	docker build -t srcd/cli-daemon:$(VERSION) -f cmd/srcd-server/Dockerfile .
+
+test-integration: clean-integration build-integration-daemon build-integration
+	$(GOTEST_INTEGRATION) github.com/src-d/engine/cmd/srcd/cmd/
--- a/Makefile
+++ b/Makefile
@@ -15,14 +15,15 @@ $(MAKEFILE):
 
 -include $(MAKEFILE)
 
-GOTEST_INTEGRATION_TAGS_LIST = integration
-GOTEST_INTEGRATION_TAGS = $(GOTEST_INTEGRATION_TAGS_LIST)
+LD_FLAGS_INTEGRATION = -X main.version="integration-testing" -X main.build=$(BUILD) -X main.commit=$(COMMIT)
+
+GOTEST_INTEGRATION_TAGS = integration
 GOTEST_INTEGRATION = $(GOTEST) -parallel 1 -count 1 -tags="$(GOTEST_INTEGRATION_TAGS)"
 
 INTEGRATION_TEST_BUILD_PATH = "build-integration"
 INTEGRATION_TEST_BIN_PATH = $(INTEGRATION_TEST_BUILD_PATH)/bin
 
-GOBUILD_INTEGRATION = $(GOCMD) build -tags "$(GOTEST_INTEGRATION_TAGS)"
+GOBUILD_INTEGRATION = $(GOCMD) build -ldflags "$(LD_FLAGS_INTEGRATION)" -tags "$(GOTEST_INTEGRATION_TAGS)"
 
 clean-integration:
 	rm -rf $(INTEGRATION_TEST_BUILD_PATH)
@@ -32,7 +33,8 @@ build-integration: GOBUILD=$(GOBUILD_INTEGRATION)
 build-integration: build
 
 build-integration-daemon:
-	docker build -t srcd/cli-daemon:$(VERSION) -f cmd/srcd-server/Dockerfile .
+	docker build --build-arg go_build_tags=integration -t srcd/cli-daemon:integration-testing -f cmd/srcd-server/Dockerfile .
 
-test-integration: clean-integration build-integration-daemon build-integration
+test-integration-no-daemon: clean-integration build-integration
 	$(GOTEST_INTEGRATION) github.com/src-d/engine/cmd/srcd/cmd/
+test-integration: build-integration-daemon test-integration-no-daemon
--- a/Makefile
+++ b/Makefile
@@ -15,15 +15,15 @@ $(MAKEFILE):
 
 -include $(MAKEFILE)
 
-LD_FLAGS_INTEGRATION = -X main.version="integration-testing" -X main.build=$(BUILD) -X main.commit=$(COMMIT)
+LD_FLAGS_IMAGE_PREFIX = -X github.com/src-d/engine/components.imageNamePrefix=integration-testing-scrd-cli-
+LD_FLAGS_INTEGRATION = -X main.version=integration-testing -X main.build=$(BUILD) -X main.commit=$(COMMIT) $(LD_FLAGS_IMAGE_PREFIX)
 
-GOTEST_INTEGRATION_TAGS = integration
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -14,3 +14,28 @@ $(MAKEFILE):
...
+INTEGRATION_TEST_BIN_PATH = $(INTEGRATION_TEST_BUILD_PATH)/bin
+
+GOBUILD_INTEGRATION = $(GOCMD) build -ldflags "$(LD_FLAGS_INTEGRATION)" -tags "$(GOTEST_INTEGRATION_TAGS)"
+
+clean-integration:
+	rm -rf $(INTEGRATION_TEST_BUILD_PATH)
+build-integration: BUILD_PATH=$(INTEGRATION_TEST_BUILD_PATH)
+build-integration: BIN_PATH=$(INTEGRATION_TEST_BIN_PATH)
+build-integration: GOBUILD=$(GOBUILD_INTEGRATION)
+build-integration: build
```
smacker: why do you need repeated target here?

cmd/srcd/cmd/init_test.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func TestInitTestSuite(t *testing.T) {
+	itt := InitTestSuite{timeout: 1 * time.Minute}
+	suite.Run(t, &itt)
+}
+
+// 1. Build docker image for daemon and name it like dev-blabla-dirty (make it automatic)
+// 2. make build
+// 3.
```
smacker: should something be here?

cmd/srcd/cmd/init_test.go
```diff
@@ -0,0 +1,205 @@
...
+func (s *InitTestSuite) SetupTest() {
+	dir, err := filepath.Abs(filepath.Join("..", "..", "..", ".integration-teststing"))
```
smacker: teststing -> testing

cmd/srcd/cmd/init_test.go
```diff
@@ -0,0 +1,205 @@
...
+	buf, err := s.runInit(s.validWorkDir)
+	require.NoError(err)
+
+	actualMsg := s.getLogMessages(buf)
+	require.Equal(5, len(actualMsg))
+
+	expectedMsg := [5]string{
+		fmt.Sprintf("removing container %s", components.Bblfshd.Name),
+		fmt.Sprintf("removing container %s", components.Daemon.Name),
+		fmt.Sprintf("removing container %s", components.Gitbase.Name),
```
smacker: are those messages always get printed in the same order?

se7entyse7en:
@smacker sorry I forget to mark as a WIP, I opened it to test whether it was working on travis. Anyway it's practically done, I just need some do some cleanup and add a description in the PR.

cmd/srcd/cmd/init_test.go
se7entyse7en: This was a personal note 😂. Gonna remove it.

Makefile
se7entyse7en: Because I want to set some env vars before running `build`. This is the only way I found, maybe there's another way. I found it [here](https://stackoverflow.com/questions/15229833/set-environment-variable-inside-target).

cmd/srcd/cmd/init_test.go
se7entyse7en: Apparently yes, the test never failed. It depends on how docker returns the names of the running containers. But this could be improved for sure.

cmd/test-utils/common.go
```diff
@@ -0,0 +1,107 @@
...
+	command := exec.CommandContext(ctx, srcdBin, args...)
+	command.Stdout = &out
+	command.Stderr = &out
+
+	return &out, command.Run()
+}
+
+var logMsgRegex = regexp.MustCompile(`.*msg="(.+)"`)
+
+func (s *IntegrationSuite) ParseLogMessages(memLog *bytes.Buffer) []string {
```
smacker: is it possible to use TraceLogMessages here? Code looks similar

se7entyse7en: Initially, that was what I wanted to do. The only thing is that `TraceLogMessages` works because it changes the format to JSON just before running `fn`. This is not possible for integration testing, unless we provide a `--log-format` flag to run `srcd` commands. That's the reason I opened #244.

Makefile
```diff
@@ -14,3 +14,28 @@ $(MAKEFILE):
 	git clone --quiet --branch $(CI_VERSION) --depth 1 $(CI_REPOSITORY) $(CI_PATH);
 
 -include $(MAKEFILE)
+
+LD_FLAGS_INTEGRATION = -X main.version="integration-testing" -X main.build=$(BUILD) -X main.commit=$(COMMIT)
+
+GOTEST_INTEGRATION_TAGS_LIST = integration
+GOTEST_INTEGRATION_TAGS = $(GOTEST_INTEGRATION_TAGS_LIST)
```
carlosms: why not `GOTEST_INTEGRATION_TAGS = integration`?

smacker: copied from lookout most probably. where we use `GOTEST_INTEGRATION_TAGS_LIST` to disable some tests on macos. Most probably not needed here

carlosms:
Did you test the windows and macos builds? I did a quick test with a PR on my fork, and both failed:
https://travis-ci.org/carlosms/engine/builds/503073554

smacker:
What do you mean? I develop on MacOs and tested it on locally of course.

If you look into your travis jobs it's failing because MacOs on travis doesn't have docker and windows doesn't have make. It's absolutely expected to fail.

Makefile
se7entyse7en: > copied from lookout most probably

Exactly 😅. Gonna remove it.

se7entyse7en:
@carlosms no, you're right, I haven't tried it on travis by removing the `if` on the `.travis.yml`. I also read that docker is not supported on OSX env in travis. So maybe we just have to remove integration testing for OSX and Windows from travis.

carlosms:
@smacker the current PR has integration tests for macos and windows enabled for master, skipping them for PRs. If merged, master will fail. What I did in my fork is enable them also for PR to see the result, but I did not add them.

@se7entyse7en I think it's fine to start with integration tests for linux only, and consider other platforms for the future.

cmd/test-utils/common.go
```diff
@@ -0,0 +1,107 @@
...
+	var out bytes.Buffer
+
+	command := exec.CommandContext(ctx, srcdBin, args...)
+	command.Stdout = &out
+	command.Stderr = &out
+
+	return &out, command.Run()
+}
+
+var logMsgRegex = regexp.MustCompile(`.*msg="(.+)"`)
```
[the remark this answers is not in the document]
se7entyse7en: @dpordomingo thank you for noticing!

smacker:
Ah. Thanks Carlos! My bad 😅

cmd/test-utils/common.go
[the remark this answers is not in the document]
se7entyse7en: Take a look at this: https://regex101.com/r/51SKuY/2.

se7entyse7en:
@dpordomingo I updated the regex with the one in the comment.
@carlosms I removed osx and windows environment from travis for integration testing.

se7entyse7en:
I've just noticed that even if I changed the names of the containers to avoid interacting with eventually running ones, there were still port conflicts. I successfully solved it with the newly added port management config feature! 🎉

se7entyse7en:
I just changed the port numbers with `default_port + 1`, with this we can at least avoid conflict with default runs.

Makefile
se7entyse7en: done

integration-testing-config.yaml
```diff
@@ -0,0 +1,15 @@
+components:
+  bblfshd:
+    port: 9433
+
+  bblfsh_web:
+    port: 8082
+
+  gitbase_web:
+    port: 8081
```
carlosms: 8081 is the default port for bblfsh_web, so this may also collide.

se7entyse7en:
@carlosms 

> One thing that I'm not so sure is adding build tags to the daemon.

To be honese I don't like it that much too. Currently it is used only for changing the container names as you correctly said. My initial solution was to use an ldflag and build it using `-X components.imagePrefix="integration-testing-srcd-cli-"`, but I wasn't able to make it work 😕.

If you prefer I can dig deeper before merging this.

carlosms:
I'm ok with the current code. But I wanted to raise my concern with going further in that direction. Because if we keep adding things then we can end up testing a special daemon image, instead of the one we release.

For me it's ok to keep it, and maybe later rethink if we can remove the build tags on the daemon when we have more tests. Personally I would not mind having the same default container names for the tests.

integration-testing-config.yaml
se7entyse7en: right 🤦‍♂️

se7entyse7en: fixed

se7entyse7en:
> But I wanted to raise my concern with going further in that direction. Because if we keep adding things then we can end up testing a special daemon image, instead of the one we release.

I agree

smacker:
I think it worths looking into why `-X` doesn't work in this case and fix it if possible.

se7entyse7en:
Ok! Gonna spend more time looking into it! 👍

cmd/test-utils/common.go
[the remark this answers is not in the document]
dpordomingo: lgtm thanks!

se7entyse7en:
Ok I found the problem for the `-X` ldflag not working. I was using the wrong path. I discovered that the correct one can be found as follows:

```
go tool nm ./build-integration/bin/srcd | grep imageNamePrefix
```

se7entyse7en:
Requested review from @carlosms again given that he was the one that initially has some concerns about build tags.

Issue #232, opened 2019-02-25
Integration tests


carlosms:
This issue is to add integration tests similar to the ones in [lookout](https://github.com/src-d/lookout/tree/master/cmd/server-test) or [gitbase](https://github.com/src-d/gitbase/blob/master/e2e/e2e_test.go). It should be as a go test suite.

Tasks:
- [x] Create a `make` rule that can build the binary and daemon, and run the integration tests
  - This should include the most basic test we can find, probably a `srcd init` call
  - The integration tests should not run as normal test with `go get`, we need to use a build tag or test flag.
  - Needs to be included to travis config
- [x] Add basic tests for each command
  - For now, let's ignore the output and check only the exit codes.
  - For the web interfaces we could at least do a wget and check it's not a 404.

se7entyse7en:
The first part is #302. After that we can split the work and parallelize maybe taking one command each.

carlosms:
One comment to consider as future improvement.
After #302 was merged I tried to run the integration tests on the released binary, skipping the build on the Makefile and downloading the bin into `build-integration/engine_linux_amd64`. The tests fail with:
```
expected: "removing container integration-testing-scrd-cli-daemon"
actual  : "removing container srcd-cli-daemon"
```

Maybe we could modify the tests to be able to run any binary, instead of the ones we build specifically for the tests. This way we can use them as kind of acceptance tests, with the released binary downloading the published docker image for srcd/cli-daemon.

smacker:
Current state:
- [x] srcd init
- [ ] srcd stop (was used in other tests, do we need separate tests for it?)
- [ ] srcd prune
- [x] srcd version
- [ ] srcd parse (maybe the command will be removed see https://github.com/src-d/feature-idea/issues/122)
- [x] srcd sql
- [ ] srcd web parse (the command will be removed see https://github.com/src-d/feature-idea/issues/122)
- [x] srcd web sql
- [x] srcd components

carlosms:
Thanks for the summary.
About parse being removed: until it's included in a relase planning, I think we should still treat it like any other command. Maybe let's not go crazy doing the most complex tests for it, but I think we should cover the basics at least.

carlosms:
When we close this one for the first "wave" of tests, we'll still have some things to cover later.
Linking here so we don't forget to open issue later: [test gitbase indexes when changing workdirs](https://github.com/src-d/engine/pull/316#pullrequestreview-213503981).

dpordomingo:
[...]

PR #314, opened 2019-03-11
Add integration tests for sql cmd

carlosms:
Part of #232.
Adds tests for `srcd sql`.

### DIFF
--- a/cmd/srcd/cmd/sql_test.go
+++ b/cmd/srcd/cmd/sql_test.go
@@ -1,0 +1,215 @@
+// +build integration
+
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	cmdtest "github.com/src-d/engine/cmd/test-utils"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type SQLTestSuite struct {
+	cmdtest.IntegrationSuite
+	testDir string
+}
+
+func TestSQLTestSuite(t *testing.T) {
+	s := SQLTestSuite{}
+	suite.Run(t, &s)
+}
+
+func (s *SQLTestSuite) SetupTest() {
+	var err error
+	s.testDir, err = ioutil.TempDir("", "sql-test")
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (s *SQLTestSuite) TearDownTest() {
+	s.RunStop(context.Background())
+	os.RemoveAll(s.testDir)
+}
+
+const showTablesOutput = `+--------------+
+|    TABLE     |
++--------------+
+| blobs        |
+| commit_blobs |
+| commit_files |
+| commit_trees |
+| commits      |
+| files        |
+| ref_commits  |
+| refs         |
+| remotes      |
+| repositories |
+| tree_entries |
++--------------+
+`
+
+func (s *SQLTestSuite) TestInit() {
+	require := s.Require()
+
+	repoPath := filepath.Join(s.testDir, "reponame")
+	err := os.Mkdir(repoPath, os.ModePerm)
+	require.NoError(err)
+
+	cmd := exec.Command("git", "init", repoPath)
+	err = cmd.Run()
+	require.NoError(err)
+
+	_, err = s.RunInit(context.TODO(), s.testDir)
+	require.NoError(err)
+
+	buf, err := s.RunSQL(context.TODO(), "select * from repositories")
+	require.NoError(err)
+
+	expected := `+---------------+
+| REPOSITORY ID |
++---------------+
+| reponame      |
++---------------+
+`
+	require.Contains(buf.String(), expected)
+}
+
+func (s *SQLTestSuite) TestNoInit() {
+	require := s.Require()
+
+	buf, err := s.RunSQL(context.TODO(), "show tables")
+	require.NoError(err)
+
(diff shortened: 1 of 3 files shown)


se7entyse7en:
CI failing.

cmd/srcd/cmd/sql_test.go
```diff
@@ -0,0 +1,209 @@
...
+	command.Stderr = &out
+	command.Stdin = &in
+
+	io.WriteString(&in, "show tables;\n")
+	io.WriteString(&in, "describe table repositories;\n")
+
+	err := command.Start()
+	require.NoError(err)
+
+	err = command.Wait()
```
smacker: could you please help me to understand how it works.
`.Wait` should block until command exits. But `repl` doesn't exit by itself and I don't see where you send stop signal to it.

cmd/srcd/cmd/sql_test.go
```diff
@@ -0,0 +1,209 @@
...
+| repositories |
+| tree_entries |
++--------------+
++---------------+------+
+|     NAME      | TYPE |
++---------------+------+
+| repository_id | TEXT |
++---------------+------+`
+
+	require.Contains(out.String(), expected)
```
smacker: I would also add test writing to stdin after we got results from the previous command. Just to simulate more real use of the repl.

cmd/srcd/cmd/sql_test.go
carlosms: Yes, this is tricky, I will add some comments to the code.

What I _think_ is happening is that the [readline lib](github.com/chzyer/readline) we use must detect that it is not in a terminal, and exits when sdtin is consumed.

You can see it if you run the bin like this. It runs the query, prints the result and exits.
```
$ echo "show tables;" | ./srcd sql
```

So we are still testing [the code that processes](https://github.com/src-d/engine/blob/master/cmd/srcd/cmd/sql.go#L108) the lines read by the repl, but without interaction. There are things we can't test, like exit on `exit;`, but at least this will fail if we break something major there.

cmd/srcd/cmd/sql_test.go
carlosms: Related to the other conversation: https://github.com/src-d/engine/pull/314#discussion_r264658396

Because the command reads stdin and exits, we are not really in interactive mode, and I think we can't test what you describe.

cmd/srcd/cmd/sql_test.go
carlosms: Added comments in 1a374ba, will squash it before merging

smacker:
Thanks for the explanation!

Most probably we still can test `exit` and other stuff if we run the command as
```
/usr/bin/bash -e "/path/to/srcd sql"
```

but better to do it in a separate PR.

cmd/srcd/cmd/sql_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	buf, err := s.RunSQL(context.TODO(), "show tables")
+	require.NoError(err)
+
+	require.Contains(buf.String(), showTablesOutput)
+}
+
+func (s *SQLTestSuite) TestValidQueries() {
+	testCases := []string{
+		"show tables",
```
dpordomingo: isn't this guaranteed by gitbase itself? If so, why replicating tests here?

cmd/srcd/cmd/sql_test.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func (s *SQLTestSuite) TestWrongQuery() {
+	testCases := []struct {
+		query string
+		err   string
+	}{
+		{
+			query: "show",
+			err:   "rpc error: code = Unknown desc = SQL query failed: Error 1105: unknown error: syntax error at position",
```
dpordomingo: do we want to ensure that error is **exactly** that? even with error number and formating?

cmd/srcd/cmd/sql_test.go
carlosms: gitbase should work for `show tables`, but we need to test that `srcd` is in fact sending the query as expected.
I put some variations to check that we don't mess with the query string. This could happen if for example tomorrow we decide to trim spaces and add a `;` to all queries, this test would detect some corner cases.

cmd/srcd/cmd/sql_test.go
carlosms: what do you propose?
My thinking here is that we want to check that the error is in fact a syntax error, not some other random crash. For example if we just checked for an error, it could happen that our client crashes with a nil pointer panic when gitbase returns an error; and the test would pass.

cmd/srcd/cmd/sql_test.go
dpordomingo: should we rewrite the query sent? should not `gitbase` to receive it as it is? why using source{d} Engine to polish sql queries?
I agree that we need to test that `srcd` sends the query, and gets the proper result, but imo it is done just sending a query, and comparing with the expected result; caring about casing, spaces or other formatting things, should be done by `gitbase`.

Anyhow, it's not a blocker to merge this PR, just a personal opinion ;)

cmd/srcd/cmd/sql_test.go
dpordomingo: I agree that we must test that it is caused a syntax error, and not a random one; what I would suggest, is not caring about the error format. Maybe just searching for `syntax error at position` would be enough because the rest is added by different layers of Engine (which do not add extra value to the user), so imo it should not be validated in tests.

carlosms: I updated the test to check only the inner error, ignoring the `rcp error` prefix. Since it's a small change, I already squashed it.

PR #316, opened 2019-03-11
Add init integration tests for workdir changes

carlosms:
Part of #232.

Adds tests for `srcd init`, adding tests that check that we are restarting gitbase when `init` is called with a different workdir, or the same workdir.

### DIFF
--- a/cmd/srcd/cmd/init_test.go
+++ b/cmd/srcd/cmd/init_test.go
@@ -6,8 +6,10 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 	"time"
@@ -31,19 +33,14 @@ func TestInitTestSuite(t *testing.T) {
 }
 
 func (s *InitTestSuite) SetupTest() {
-	dir, err := filepath.Abs(filepath.Join("..", "..", "..", ".integration-testing"))
+	var err error
+	s.testDir, err = ioutil.TempDir("", "init-test")
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	s.testDir = dir
 	s.validWorkDir = filepath.Join(s.testDir, "valid-workdir")
 	s.invalidWorkDir = filepath.Join(s.testDir, "invalid-workdir")
-
-	err = os.Mkdir(s.testDir, os.ModePerm)
-	if err != nil {
-		log.Fatal(err)
-	}
 
 	err = os.MkdirAll(s.validWorkDir, os.ModePerm)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/srcd/cmd/init_test.go
```diff
@@ -31,20 +33,15 @@ func TestInitTestSuite(t *testing.T) {
 }
 
 func (s *InitTestSuite) SetupTest() {
-	dir, err := filepath.Abs(filepath.Join("..", "..", "..", ".integration-testing"))
+	var err error
+	s.testDir, err = ioutil.TempDir("", "init-test")
```
smacker: 👍

se7entyse7en: I'm gonna try this first locally coz if I remember well I also did like this initially, but then the folder was created in a path that docker wasn't able to mount (on OSX).

carlosms:
Rebased to see if tests pass as they should.
I applied both your suggestions, already squashed in 97bec04.

cmd/srcd/cmd/init_test.go
se7entyse7en: Yup, unfortunately it has some problems on OSX:

Screenshot 2019-03-12 at 14 23 59

In my case:

```
✘  se7entyse7en in ~/Projects/.go-workspace/src/github.com/src-d/engine (carlosms-integration-init)  $ echo $TMPDIR                                                              [03/12/19 | 14:23:46.790]
/var/folders/80/4kkg57_13jl8wyslvlzdfhr80000gn/T/
```

If we want to use `ioutil.TempDir` maybe we could set the env var to another location.

se7entyse7en:
Great tests! 👍

cmd/srcd/cmd/init_test.go
carlosms: Is it enough to do this in our makefile?
```
TMPDIR="./integration-test-tmp"

test-integration-no-build:
	rm -rf $(TMPDIR)
	mkdir $(TMPDIR)
	$(GOTEST_INTEGRATION) github.com/src-d/engine/cmd/srcd/cmd/

test-integration: clean build docker-build test-integration-no-build
```

dpordomingo: coud we add the `rm -rf $(TMPDIR)` to the `clean` target?, and use it as `test-integration-no-build` dependency?

cmd/srcd/cmd/init_test.go
```diff
@@ -200,3 +197,104 @@ func (s *InitTestSuite) TestWithRunningOtherComponents() {
...
 	}
 }
+
+func (s *InitTestSuite) initGitRepo(path string) {
+	s.T().Helper()
+
+	err := os.MkdirAll(path, os.ModePerm)
+	s.Require().NoError(err)
+
+	cmd := exec.Command("git", "init", path)
```
dpordomingo: should we ensure someway that `git` is already available?

cmd/srcd/cmd/init_test.go
se7entyse7en:
```test-integration-no-daemon: clean-integration build-integration
	TMPDIR=`pwd`/integration-test-tmp
	rm -rf $(TMPDIR)
	mkdir $(TMPDIR)
	$(GOTEST_INTEGRATION) github.com/src-d/engine/cmd/srcd/cmd/
```

se7entyse7en: The problem is that we should restore the original value of `TMPDIR`, as the user doesn't expect it to be changed.

dpordomingo: I saw `$(PWD)` and `$(shell pwd)`, but never `'pwd'`
I wonder if that's what is causing your problem @se7entyse7en

se7entyse7en: @dpordomingo you're right! This worked and didn't override the env var once it finished. 👍 

```TMPDIR=$(PWD)/integration-test-tmp

test-integration-no-daemon: clean-integration build-integration
	rm -rf "$(TMPDIR)"
	mkdir "$(TMPDIR)"
	$(GOTEST_INTEGRATION) github.com/src-d/engine/cmd/srcd/cmd/
```

carlosms: Since it's not straightforward, and we already have some test in master using `ioutil.TempDir`, would you agree to keep this PR working for linux, and find a proper and clean solution in a different PR later? If you agree I'll open a new issue with a link to this discussion.

carlosms:
> should we ensure in these tests that indexes are properly handled when changing between directories? or is that guaranteed by gitbase?

You are right, we should test that. There is a tricky issue hidden there, and is that we are using a workdir in /tmp, but the datadir is always `~/.srcd/gitbase`. Ideally we should be using a clean dir for the tests also.

I think it makes sense to leave it as future improvement, I left a comment in the main issue here: https://github.com/src-d/engine/issues/232#issuecomment-472127062

cmd/srcd/cmd/init_test.go
carlosms: If `git` is not installed the tests will complain in the next `s.Require().NoError(err)`.
What else did you have in mind?

cmd/srcd/cmd/init_test.go
se7entyse7en: it's ok for me

cmd/srcd/cmd/init_test.go
dpordomingo: maybe as a precondition for the `test-integration` make target...
but not a strong opinion.

carlosms: I understand your point, but I'd prefer to keep it simple, at least for now. We could check many things (docker is installed, etc), but in the end the tests are meant for our CI, that we control, and the developers machines. So to me it's safe to assume some tools.

dpordomingo: is the computer on? do we have network? :dancer:

Issue #284, opened 2019-03-05
web sql should wait for gitbase to be ready before launching gitbase-web


carlosms:
For a `srcd sql "query"` we wait until gitbase is ready. We should do the same for `srcd web sql`. Otherwise gitbase-web opens, but queries fail with an "internal error" message.

PR #318, opened 2019-03-12
make web sql wait for gitbase to be ready

smacker:
Fix: #284 

Output:
```
$ srcd web sql
INFO[0003] this is taking a while, it might take a few more minutes while we install all the required images
INFO[0019] waiting for gitbase to be ready, done
INFO[0022] this is taking a while, if this is the first time you launch this web client, it might take a few more minutes while we install all the required images
```

It repeats "this is taking a while" if there are no images. Do you think is it a big problem?
I don't see how to avoid it without rewriting a lot.


dpordomingo:
I don't see a problem here ;)

Issue #291, opened 2019-03-06
Standarize error messages


dpordomingo:
discovered at https://github.com/src-d/empathy-sessions/issues/37#issuecomment-470037848

Errors do not look like to follow a fixed format, here are some examples:

### Errors From Port Conflicts

```shell
$ srcd sql "SHOW tables;"
FATA[0000] could not start gitbase: rpc error: code = Unknown desc = could not create srcd-cli-bblfshd:
could not start container: srcd-cli-bblfshd: Error response from daemon:
driver failed programming external connectivity on endpoint srcd-cli-bblfshd (3120...):
Bind for 0.0.0.0:9432 failed: port is already allocated 
```

```shell
$ srcd sql "SHOW tables;"
FATA[0000] could not start gitbase: rpc error: code = Unknown desc = could not create srcd-cli-gitbase:
could not start container: srcd-cli-gitbase: Error response from daemon:
driver failed programming external connectivity on endpoint srcd-cli-gitbase (5e3b...):
Error starting userland proxy: listen tcp 0.0.0.0:3306: bind: address already in use 
```

This one is funny: it outputs the same error log twice, but also the `srcd parse uast --help` message in the middle :)

```shell
$ srcd parse uast examples/example.php
INFO[0000] detected language: php             
Error: could not list drivers: rpc error: code = Unknown desc = could not create srcd-cli-bblfshd:
could not start container: srcd-cli-bblfshd: Error response from daemon:
driver failed programming external connectivity on endpoint srcd-cli-bblfshd (fc8a...):
Bind for 0.0.0.0:9432 failed: port is already allocated
Usage:
  srcd parse uast [file-path] [flags]

...

could not list drivers: rpc error: code = Unknown desc = could not create srcd-cli-bblfshd:
could not start container: srcd-cli-bblfshd: Error response from daemon:
driver failed programming external connectivity on endpoint srcd-cli-bblfshd (fc8a...):
Bind for 0.0.0.0:9432 failed: port is already allocated
```

### Error When RPC Message Is Too Big

```shell
$ srcd sql "SELECT cf.file_path, f.blob_content
FROM ref_commits r
NATURAL JOIN commit_files cf
NATURAL JOIN files f
WHERE r.ref_name = 'HEAD'
AND r.history_index = 0;"

2019/03/04 10:55:11 rpc error: code = ResourceExhausted desc = grpc:
                    received message larger than max (5175028 vs. 4194304)
```

smacker:
Proposal:
[...]

PR #319, opened 2019-03-12
Shows supported languages when an unknown one is requested

### LINKED ISSUE #294: Parsing files logs when using unknown lang
discovered at https://github.com/src-d/empathy-sessions/issues/37#issuecomment-470042245

If you choose a nonexistent driver, it outputs the help message, but it should list the available languages instead.

```shell
$ srcd parse uast examples/example.php --lang unknown
Error: language unknown is not supported
Usage:
  srcd parse uast [file-path] [flags]
...
```

se7entyse7en:
Closes #294.

### DIFF
--- a/cmd/srcd/cmd/parse.go
+++ b/cmd/srcd/cmd/parse.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -95,8 +96,9 @@ The remaining nodes are printed to standard output in JSON format.`,
 			return fmt.Errorf("could not list drivers: %v", err)
 		}
 
-		if !isSupportedLanguage(resp.Drivers, lang) {
-			return fmt.Errorf("language %s is not supported", lang)
+		err = checkSupportedLanguage(resp.Drivers, lang)
+		if err != nil {
+			return err
 		}
 
 		stream, err := c.ParseWithLogs(ctx, &api.ParseRequest{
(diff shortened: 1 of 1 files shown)


se7entyse7en:
A couple of things. Do we want to only show the supported languages or do we also want to avoid showing the usage message?

I noticed that there's a `SilenceUsage` flag that avoids printing the usage message. We could also eventually set the usage depending on the error that we encounter, but honestly, I don't like this approach as it seems a bit hacky.

Another thing that I noticed is that the error is actually printed twice, are we ok with that?

Screenshot 2019-03-12 at 23 20 44

We could eventually hide the first one using the `SilenceErrors` flag.

Here how the error appears depending on these two above-mentioned flags:

1. `SilenceUsage: true` and `SilenceErrors: false`

    Screenshot 2019-03-12 at 23 23 25

2. `SilenceUsage: false` and `SilenceErrors: true`

    Screenshot 2019-03-12 at 23 24 09

3. `SilenceUsage: true` and `SilenceErrors: true`

    Screenshot 2019-03-12 at 23 24 48

se7entyse7en:
I've just noticed that this is kind of related to https://github.com/src-d/engine/issues/291.

cmd/srcd/cmd/parse.go
```diff
@@ -95,8 +96,11 @@ The remaining nodes are printed to standard output in JSON format.`,
 			return fmt.Errorf("could not list drivers: %v", err)
 		}
 
-		if !isSupportedLanguage(resp.Drivers, lang) {
-			return fmt.Errorf("language %s is not supported", lang)
+		langaugesSuppInfo := getLanguagesSupportInfo(resp.Drivers, lang)
```
carlosms:
```suggestion
		languagesSuppInfo := getLanguagesSupportInfo(resp.Drivers, lang)
```

cmd/srcd/cmd/parse.go
```diff
@@ -216,12 +220,25 @@ func parseLang(ctx context.Context, client api.EngineClient, path string, b []by
...
 }
 
-func isSupportedLanguage(supportedDrivers []*api.ListDriversResponse_DriverInfo, desired string) bool {
+type languageSupportInfo struct {
+	supported   []string
+	requested   string
+	isSupported bool
+}
+
+func getLanguagesSupportInfo(supportedDrivers []*api.ListDriversResponse_DriverInfo, desired string) *languageSupportInfo {
```
carlosms: For some reason this feels odd. The struct return type is there to provide the info you need to build the error message, only in case of error.
Instead of that, we could build the nice error message inside this function. Something like:
```go
func checkSupportedLanguage(supportedDrivers []*api.ListDriversResponse_DriverInfo, desired string) error
```

smacker:
Let's keep duplicated error for this PR. It's how all other errors work right now.
And fix it in another PR. (from the screenshots above I prefer `SilenceUsage: false and SilenceErrors: true` but maybe it's not the best output in some other cases)

cmd/srcd/cmd/parse.go
se7entyse7en: done, I already squashed it as it is a very small thing.

se7entyse7en:
Later I'm gonna create the issue for [this](https://github.com/src-d/engine/pull/319#issuecomment-472383336).

Issue #252, opened 2019-03-04
Warning about version


smacker:
This `WARN` is returned always.
What does it mean? 
What can I do as a user?

```shell
$ srcd init /projects/src/github.com/torvalds
WARN[0001] unable to list the available daemon versions on Docker Hub: 
can't find compatible image in docker registry for srcd/cli-daemon 
```

![image](https://user-images.githubusercontent.com/406916/53729949-e032cf00-3e76-11e9-8c5a-5dba0c075b00.png)

carlosms:
This probably happens because the release is `-rc`

smacker:
Yep. Maybe we should add support for `-rc` versions.

Issue #308, opened 2019-03-06
Update command


se7entyse7en:
From [here](https://github.com/src-d/empathy-sessions/issues/32).

> during the empathy session a new version of engine has been released. It would be very useful to have a command `srcd update` that detects the current installation directory, downloads the new version and moves it in the same location.

smacker:
More ideas from Marvin: https://github.com/src-d/engine/pull/323#issuecomment-472757370

vmarkovtsev:
I strongly +1 this as a casual user. Please add this!

PR #323, opened 2019-03-13
add support for pre releases

smacker:
Fix: #252

For pre-release client version it will always return exact the same
version of the server due to unstable nature of pre-releases and
possible breaking changes between them or the next stable version.

Signed-off-by: Maxim Sukharev <max@smacker.ru>

### DIFF
--- a/docker/dockerhub.go
+++ b/docker/dockerhub.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/blang/semver"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


carlosms:
LGTM.
But this makes me wonder... If you have a client 0.11.0 and we release 0.11.1, the client will silently use the daemon image 0.11.1, and it will not notify the user of a new release.
But what if 0.11.1 was released because we found an important bug in the `srcd` itself?

smacker:
if it is really important we can release 0.12.0. Though auto-update for srcd-cli would be nice.
We have this issue: https://github.com/src-d/engine/issues/308
When it's implemented we can just run automatically for patch versions.

Another option is to notify user always about new versions. But it might be a bit annoying if we continue to follow "release often" strategy.

se7entyse7en:
We can use the config file to let the user decide whether to enable auto-updates or not by exposing some updates policy such as:

- `enable_autoupdate`: `true`, `false`
- `update_policy`: `"always"`, `"minors"`, `"patches only"`,
- `notify_update_available`: `true`, `false`
- `notify_update_occurs`: `true`, `false`

These should cover each case.

For example zsh has the following env vars:
```
# Uncomment the following line to disable bi-weekly auto-update checks.
# DISABLE_AUTO_UPDATE="true"

# Uncomment the following line to change how often to auto-update (in days).
# export UPDATE_ZSH_DAYS=13
```

But in our case I think that the number of days doesn't tell anything.

smacker:
I'm not sure we can do auto-updates for minor versions while we are at 0.x.x.
Minor means breaking change so client & server have to be the same version which means while we update cli we would have to restart server too. Which might be unexpected for a user especially if they run some query or whatever.
Auto-update for patch version is safe. Cli should be able to work with any patch version of the server and the server will be updated automatically when user does `init` (or anything else that causes init)

PR #324, opened 2019-03-13
Add integration tests for parse cmd

carlosms:
Part of #232.
Tests for `srcd parse drivers`, `srcd parse lang`, `srcd parse uast`

### DIFF
--- a/cmd/srcd/cmd/parse_test.go
+++ b/cmd/srcd/cmd/parse_test.go
@@ -1,0 +1,222 @@
+// +build integration
+
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	cmdtest "github.com/src-d/engine/cmd/test-utils"
+
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+type ParseTestSuite struct {
+	cmdtest.IntegrationSuite
+	testDir string
+}
+
+func TestParseTestSuite(t *testing.T) {
+	s := ParseTestSuite{}
+	suite.Run(t, &s)
+}
+
+type testCase struct {
+	path     string
+	filename string
+	lang     string
+}
+
+// Uses files from https://github.com/leachim6/hello-world
+var testCases = []testCase{
+	{
+		path:     filepath.FromSlash("testdata/hello.py"),
+		filename: "hello.py",
+		lang:     "python",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello-py3.py"),
+		filename: "hello-py3.py",
+		lang:     "python",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.cpp"),
+		filename: "hello.cpp",
+		lang:     "c++",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.java"),
+		filename: "hello.java",
+		lang:     "java",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.js"),
+		filename: "hello.js",
+		lang:     "javascript",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.bash"),
+		filename: "hello.bash",
+		lang:     "shell",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.rb"),
+		filename: "hello.rb",
+		lang:     "ruby",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.go"),
+		filename: "hello.go",
+		lang:     "go",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.cs"),
+		filename: "hello.cs",
+		lang:     "c#",
+	},
+	{
+		path:     filepath.FromSlash("testdata/hello.php"),
+		filename: "hello.php",
+		lang:     "php",
+	},
+}
+
+func (s *ParseTestSuite) SetupTest() {
+}
+
+func (s *ParseTestSuite) TearDownTest() {
+	s.RunStop(context.Background())
+}
+
+func (s *ParseTestSuite) TestDriversList() {
+	require := s.Require()
+
+	out, err := s.RunCommand(context.TODO(), "parse", "drivers", "list")
+	outStr := out.String()
+
+	require.NoError(err, outStr)
+
+	/* Example output:
+
+	LANGUAGE	VERSION
+	----------	----------
+	python		v2.8.0
+	cpp		v1.1.0
+	java		v2.5.0
+	javascript	v2.6.0
+	bash		v2.4.0
+	ruby		v2.9.0
+	go		v2.5.0
+	csharp		v1.4.0
+	php		v2.7.0
+	*/
+
+	// Simple checks to see if it's the table, and contains a known driver
+	expected := regexp.MustCompile(`LANGUAGE\s+VERSION`)
+	require.Regexp(expected, outStr)
+	expected = regexp.MustCompile(`javascript\s+v\S+`)
+	require.Regexp(expected, outStr)
+}
+
+func (s *ParseTestSuite) TestLang() {
+	for _, tc := range testCases {
+		s.T().Run(tc.filename, func(t *testing.T) {
+			require := require.New(t)
+
+			// Check the language is detected
+			out, err := s.RunCommand(context.TODO(), "parse", "lang", tc.path)
+			require.NoError(err, out.String())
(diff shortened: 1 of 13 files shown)


cmd/srcd/cmd/parse_test.go
```diff
@@ -0,0 +1,257 @@
...
+	bash		v2.4.0
+	ruby		v2.9.0
+	go		v2.5.0
+	csharp		v1.4.0
+	php		v2.7.0
+	*/
+
+	// Simple checks to see if it's the table, and contains a known driver
+	require.Contains(outStr, "LANGUAGE	VERSION")
+	expected := regexp.MustCompile(`javascript	v\S+`)
```
smacker: better `javascript\s+v\S+` because I'm not sure that current regexp will still be valid if there is any new driver with a longer name.

cmd/srcd/cmd/parse_test.go
```diff
@@ -0,0 +1,257 @@
...
+				t.Run(mode, func(t *testing.T) {
+
+					for _, lang := range langOpts {
+						t.Run(lang, func(t *testing.T) {
+
+							for _, query := range queryOpts {
+								t.Run(query, func(t *testing.T) {
+
+									require := require.New(t)
+
```
smacker: wow. The cyclomatic complexity of this function must be crazy.

cmd/srcd/cmd/parse_test.go
```diff
@@ -0,0 +1,257 @@
...
+									// RunCommand mixes stdout and stderr. To parse the UAST output properly
+									// we need to read stdout only
+									uastOut, err := s.runCommandStdout(context.TODO(), "parse", args...)
+
+									// ----------------
+									// TODO Temporary test skip, it fails for cpp, bash, and csharp.
+									// See https://github.com/src-d/engine/issues/297
+									if err != nil && strings.Contains(uastOut.String(), "is not supported") {
+										t.Skip("TEST FAILURE IS A KNOWN ISSUE (#297): " + uastOut.String())
```
smacker: I would better `if` only for failing languages. So we would still test no-error for the rest.

carlosms: You mean like this?
```go
if err != nil &&
	(tc.lang == "c++" || tc.lang == "shell" || tc.lang == "c#") &&
	strings.Contains(uastOut.String(), "is not supported") {

	t.Skip("TEST FAILURE IS A KNOWN ISSUE (#297): " + uastOut.String())
}
```

cmd/srcd/cmd/parse_test.go
carlosms: And the code is super ugly, the level of indentation alone is an eye sore. But... it's for the tests, and only to mix all the table test cases.
Should we try to improve this somehow?

cmd/srcd/cmd/parse_test.go
carlosms: Applied and squashed

cmd/srcd/cmd/parse_test.go
smacker: something like:
```go
if lang == "no-lang-option" && tc.lang == "c++" || tc.lang == "shell" || tc.lang == "c#" {
  t.Error(err) // so we know when it's fixed and can remove skip
  t.Skip("TEST FAILURE IS A KNOWN ISSUE (#297)")
}
```

the main problem currently if python, for example, returns an error it will be skipped and CI would return success.

cmd/srcd/cmd/parse_test.go
smacker: to be clear: I propose to rewrite it to avoid so many inner for loops. It's hard to read.

smacker: without removing for loops themselves but at least separating it from actual tests we can do something like:
```go
func Test() {
  expandedTestCases := []expandedTestCase{}
  for _, tc := range testCases {
	for _, mode := range modes {
		for _, lang := range langOpts {
			for _, query := range queryOpts {
				name := ""
				expandedTestCases = append(expandedTestCases, expandedTestCase{...})
			}
		}
	}
  }

  for _ tc := range expandedTestCases {
	  // actual test
  }
}

type expandedTestCase struct {
	testCase

	name string
	mode string
	lang string
	query string
}
```

smacker: another option is keep this for loops with t.Run as is but move "meat" to separate function.

carlosms: I gave a try in 41aadd2, I see I found a solution similar to your first proposal.

cmd/srcd/cmd/parse_test.go
carlosms: Changed in 7fb3fab

cmd/srcd/cmd/parse_test.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+type testCase struct {
+	url      string
+	filename string
+	lang     string
+}
+
+// Uses files from https://github.com/leachim6/hello-world
+var testCases = []testCase{
```
se7entyse7en: Isn't it better to have these files locally in a fixture dir?

dpordomingo: yes please :pray:

carlosms: It's funny how obvious it sounds now that you mention it.

Done in 990247d

se7entyse7en: 😂

se7entyse7en: LGTM 👍, thank you

cmd/srcd/cmd/parse_test.go
dpordomingo: I agree that readability is important.
I'd purpose a cleaner way to combine the arguments (permutations is about ordering, a different thing)
https://play.golang.org/p/i991wbz9WaS
imo, that way is clear that you're combining queries, modes, and langs, per test case.

I'd also extract the generation of the command arguments from the `TestUast` test case

cmd/srcd/cmd/parse_test.go
dpordomingo: better :)

cmd/srcd/cmd/parse_test.go
```diff
@@ -0,0 +1,227 @@
...
+			// Check the language is detected
+			out, err := s.RunCommand(context.TODO(), "parse", "lang", tc.path)
+			require.NoError(err, out.String())
+			require.Equal(tc.lang+"\n", out.String())
+		})
+	}
+}
+
+// same as RunCommand, but captures only stdout instead of stdout + stderr
+func (s *ParseTestSuite) runCommandStdout(ctx context.Context, cmd string, args ...string) (*bytes.Buffer, error) {
```
dpordomingo: ouch :cry: 
Why not moving it parallel to the other one [IntegrationSuite::RunCommand](https://github.com/src-d/engine/blob/master/cmd/test-utils/common.go#L33), and making it public?
And even more, I'd reconsider the necessity of merging the `stderr` and `stdout`, because not doing it it could be easily ensured what is in the `stderr` or in the `stdout`, or even we could ensure that a command does not write in the `stderr`

smacker:
In go it's better to use directory name `testdata`. 
You can read more about it here for example: https://medium.com/@povilasve/go-advanced-tips-tricks-a872503ac859

stdlib uses it and it has nice feature of always resolving correctly.

cmd/srcd/cmd/parse_test.go
smacker: My 0.0000001 cents: I also noticed this, but I think it's better to improve in another PR (maybe separate stderr/out for existing tests too)

dpordomingo: agree on doing it in a separated issue if you prefer to handle it that way @carlosms

carlosms:
:+1: I didn't know about testdata. The dir is relative to the test file just like before, but it's very nice that go ignores this dir, so we don't have to worry about go file fixtures.
Changed in ad11bf5.

cmd/srcd/cmd/parse_test.go
carlosms: ok, done in 5b68fd3

cmd/srcd/cmd/parse_test.go
carlosms: yes, there are things that can be refactored after this first set of tests is done. There is also some repetition in the tmp dir initialization, and I'm sure some other things will come up. Let's take care of this in a separate PR.

dpordomingo: I'd love you if you also open the issue. 😘

carlosms: Done in #347

Issue #258, opened 2019-03-04
Port conflict error is not very user friendly


smacker:
It's also hard to find information how to fix conflict (using config file)

```shell
$ srcd sql "SHOW tables;"
FATA[0000] could not start gitbase: rpc error: code = Unknown desc = could not create srcd-cli-bblfshd:
could not start container: srcd-cli-bblfshd: Error response from daemon:
driver failed programming external connectivity on endpoint srcd-cli-bblfshd (8545...):
Bind for 0.0.0.0:9432 failed: port is already allocated 
```

Screenshot from the empathy session log:
![image](https://user-images.githubusercontent.com/406916/53733741-36f1d600-3e82-11e9-8cd4-40e555ad3cce.png)

carlosms:
Screenshot from the empathy session log:
https://user-images.githubusercontent.com/406916/53733741-36f1d600-3e82-11e9-8cd4-40e555ad3cce.png

dpordomingo:
I also found this issue

I'd expect:

```shell
$ srcd sql "SHOW tables;"
FATAL[0000] could not start gitbase. Port 9432 is already allocated.
You can define the port to be binded by ${service} in ${confFile} under ${portName}, and then run:
'srcd init ${workdir} --config ${confFile}'
```
where:
- `${workdir}` is `/path/to/workdir`
- `${confFile}` is `~/.srcd/config.yml`
- `${service}` is `srcd-cli-bblfshd`
- `${portName}` is `components.bblfshd.port`

PR #328, opened 2019-03-14
Better port conflict error

smacker:
Fix: #258

To fix the issue I introduced 2 new concepts:
1. `docker.Error` - docker api doesn't export almost any errors. To distinguish what has happened I added functions to parse error message and generate our own typed errors. Currently, it can recognize only error coming from docker daemon and bind error. Imo, it's better to keep such code in one place instead of hacking all around codebase. And if in the future docker api will start to provide more information we will be able to change only this module.
2. `cli.fatal` - replacement for `logrus.Fatal` which we use right now to notify a user about an error. This function converts `docker.Error`s (or any other types if we need any) to a human-friendly message.

Current output for bind error:
```
$ go run cmd/srcd/main.go sql "SHOW tables;"
FATA[0000] could not start gitbase: Port 9432 is already allocated.
You can define the port to be bound by srcd-cli-bblfshd in path/to/config.yml, and then run:
srcd init [workdir] --config path/to/config.yml
exit status 1
```


smacker:
One more note: in some cases instead of `logrus.Fatal` we return error and rely on cobra to print the message. There is another issue about unifying it. It's easy to change `fatal` function to return error if we want to use cobra everywhere. But for this issue, I think it's out of the scope.

cmd/srcd/cmd/errors.go
```diff
@@ -0,0 +1,29 @@
...
+	err = docker.ParseErr(err)
+	errString := err.Error()
+
+	switch e := err.(type) {
+	case *docker.ContainerBindErr:
+		// TODO(max): instead of placeholders we can actually ask daemon for current config if we would have such API
+		confFile := "path/to/config.yml"
+		workdir := "[workdir]"
+
+		errString = "Port " + e.Port + " is already allocated.\n" +
```
carlosms: Maybe we could use the default path for the config file ($HOME/.srcd/config.yml).
Also I think we should link to https://docs.sourced.tech/engine/learn-more/commands#srcd, otherwise as a user you don't have a way to know what you are supposed to write in the file.

cmd/srcd/cmd/stop.go
```diff
@@ -11,7 +10,7 @@ var stopCmd = &cobra.Command{
 	Short: "Stops all containers.",
 	Run: func(cmd *cobra.Command, args []string) {
 		if err := components.Stop(); err != nil {
-			logrus.Fatal(err)
+			fatal(err, "count not stop containers")
```
se7entyse7en: typo

cmd/srcd/cmd/prune.go
```diff
@@ -13,7 +12,7 @@ var pruneCmd = &cobra.Command{
 		withImages, _ := cmd.Flags().GetBool("with-images")
 
 		if err := components.Prune(withImages); err != nil {
-			logrus.Fatal(err)
+			fatal(err, "cound not prune components")
```
se7entyse7en: typo

smacker:
applied comments in separate commits. push-force due to rebase on master.

cmd/srcd/cmd/errors.go
dpordomingo: I'd avoid linking external resources in go files if we don't have a way to validate links integrity.
I agree on mentioning  default place to store it.
What I'd really really really like, is a command to generate that file. Wouldn't be great? 😅

PR #329, opened 2019-03-15
Adds a Known Issues section in the README

### LINKED ISSUE #156: Let user know that on macOS gitbase works MUCH slower with engine
The same simple query in engine:

```
21 rows in set (2 min 9.63 sec)
```

native:

```
21 rows in set (3.69 sec)
```

According to my Activity Monitor it happens because of FS mapping into docker.

se7entyse7en:
Closes #156.

I decided to add a reference to Gitbase to give more context as it is already cited in the architecture section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ You can find a version of this docummentation, properly rendered, at https://doc
 - [Architecture](#architecture)
 - [Babelfish UAST](#babelfish-uast)
 - [Clients & Connectors](#clients-and-connectors)
+- [Known Issues](#known-issues)
 - [Community](#community)
 - [Contributing](#contributing)
 - [Credits](#credits)
@@ -66,7 +67,7 @@ Follow these instructions:
 
 #### 2. Install source{d} Engine
 
-Download the **[latest release](https://github.com/src-d/engine/releases)** for MacOS (Darwin), Linux or Windows. *The support for Windows is currently experimental.*
+Download the **[latest release](https://github.com/src-d/engine/releases)** for MacOS (Darwin), Linux or Windows. *The support for Windows is currently experimental.* Please take also some time to have a look at the [Known Issues](#known-issues) section.
 
 **MacOS:**
 
@@ -336,6 +337,18 @@ srcd parse drivers install python
 ```
 
 To see which languages are available, check the table of [supported languages](https://docs.sourced.tech/babelfish/languages).
+
+
+## Known Issues
+
+This section contains a list of some known issues and how they can be addressed:
+
+- Running source{d} Engine on macOS has some performance issues:
+
+    SQL queries on MacOS will run slower. If your workdir contains many repositories, then the underlying gitbase component (see the [Architecture](#architecture) section) other than running queries slower, may even fail to initialize. The problem is related to the way Docker mounts volumes in macOS. See [#156](https://github.com/src-d/engine/issues/156) for more context.
+
+    If you plan to run source{d} Engine on a big dataset we suggest you to do it on a Linux system.
+
 
 ## Clients and Connectors
 


README.md
```diff
@@ -337,6 +338,18 @@ srcd parse drivers install python
...
 To see which languages are available, check the table of [supported languages](https://docs.sourced.tech/babelfish/languages).
 
+
+## Known Issues
+
+This section contains a list of some known issues and how they can be addressed:
+
+- Running source{d} Engine on MacOS has some performance issues:
+
+    SQL queries on MacOS will run slower. If your workdir contains many repositories, then the underlying Gitbase component (see the [Architecture](#architecture) section) other then running queries slower, may even fail to initialize. The problem seems to be related to FS mapping into Docker for MacOS. Some context can be found here [#156](https://github.com/src-d/engine/issues/156).
```
carlosms:
```suggestion
    SQL queries on MacOS will run slower. If your workdir contains many repositories, then the underlying gitbase component (see the [Architecture](#architecture) section) other than running queries slower, may even fail to initialize. The problem is related to the way Docker mounts volumes in macOS. See [#156](https://github.com/src-d/engine/issues/156) for more context.
```

README.md
```diff
@@ -337,6 +338,18 @@ srcd parse drivers install python
...
 To see which languages are available, check the table of [supported languages](https://docs.sourced.tech/babelfish/languages).
 
+
+## Known Issues
+
+This section contains a list of some known issues and how they can be addressed:
+
+- Running source{d} Engine on macOS has some performance issues:
+
+    SQL queries on MacOS will run slower. If your workdir contains many repositories, then the underlying gitbase component (see the [Architecture](#architecture) section) other than running queries slower, may even fail to initialize. The problem is related to the way Docker mounts volumes in macOS. See [#156](https://github.com/src-d/engine/issues/156) for more context.
```
dpordomingo: Can gitbase _evan fail to initialize_?
when did we saw this problem? can we describe it better? maybe providing a sort of approximation of what "a big dataset" is?

se7entyse7en: > Can gitbase evan fail to initialize?

Maybe @smacker can help on this as I just reported what he said [here](https://github.com/src-d/engine/issues/156#issuecomment-469325407).

> maybe providing a sort of approximation of what "a big dataset" is

That's actually a big point. Without knowing the gitbase internals I actually don't know what to say, how much it depends on the number of repos, or size of the repos or whatever. Maybe @src-d/data-retrieval could help on this.

mcarmonaa: I guess you were  looking for the help from the new awesome @src-d/data-processing team instead of @src-d/data-retrieval, gitbase is not anymore a data-retrieval's project.

se7entyse7en: Thank you @mcarmonaa.

smacker:
does it make sense to add this known issue if we found [the way](https://github.com/src-d/engine/issues/330) to fix it? (though I think we need to test it more thoroughly, we could repeat one of the tasks @gomesfernanda did and suffered with recently)

se7entyse7en:
> does it make sense to add this known issue if we found the way to fix it?

I don't know if it is actually a fix, it's surely a way to mitigate the problem. IMO once we change the consistency of the mounts we still have to inform the user about the fact that queries will run slower on OSX. But maybe "issue" will no longer be the most appropriate term given that it actually depends on the fs mapping.

> though I think we need to test it more thoroughly

I agree. So far we've just acknowledged that the issue exists, but we don't know exactly for example if some type of queries are more affected than others (the user may want to know which kind of queries are more affected). Additionally, the user may want to know how much slower the queries are. Different users may consider an +x% in response time acceptable and others not.

se7entyse7en:
> From the discussion above I see there is no a clear understanding of how, when an why the problem appears.

Well actually the *how* is that the queries run slower, the *when* is when using OSX (yes it's not clear if for every query, but I *guess* so), the *why* is most *probably* due to the FS mapping and it makes sense given that [changing the consistency](https://github.com/src-d/engine/issues/156#issuecomment-473252486) improved considerably the performances at least for that query.

> Imo we should be able to describe better the problem and confirm that there is nothing to be done in our side. If we get that point, I think we should open an issue in gitbase with proper description and
steps to reproduce the issue, and then link it here.

I agree that the problem should be described better. I'm ok in opening an issue in gitbase about queries running slower on OSX when dockerized. In the meantime, from our side, IMO fixing [this](https://github.com/src-d/engine/issues/330) could be done without any particular issue.

smacker:
we don't need it anymore. Thanks anyway!

PR #339, opened 2019-03-15
Add integration tests for web cmd

carlosms:
Part of #232.
Tests for `web sql`, `web parse`.

### DIFF
--- a/cmd/srcd/cmd/web_test.go
+++ b/cmd/srcd/cmd/web_test.go
@@ -1,0 +1,158 @@
+// +build integration
+
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/blang/semver"
+	cmdtest "github.com/src-d/engine/cmd/test-utils"
+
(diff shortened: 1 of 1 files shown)


cmd/srcd/cmd/web_test.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (s *WebTestSuite) testCommon(subcmd string, assertions func(url string)) {
+	require := s.Require()
+
+	var out bytes.Buffer
+
+	command := s.CommandContext(context.TODO(), "web", subcmd)
+	command.Stdout = &out
+	command.Stderr = &out
```
dpordomingo: why not writing `stderr` into a new error buffer, and then assert that if there were no errors, there is nothing written in the error buffer?

carlosms: We are sending the log messages to `stderr`. We could simply ignore `stderr` and capture `stdout` only, but I left it combined to work in a consistent way with `RunCommand`.

dpordomingo: errors in `stdout` :(
ok...
given that constraint, consistency was a good decision :+1:

carlosms: No, error messages also go to `stderr`.

Issue #345, opened 2019-03-17
Engine package: MaxAllowedPacket set to 65536 (wanted 32M)


jamesnaftel:
In engine/cmd/srcd-server/engine/sql.go, the MaxAllowedPacket is set to 65536.  Watching "justforfunc live: hacking on a CLI with cobra, gRPC, and Docker", @campoy said he wanted the packet size to be 32M.  
`
	cfg := mysql.Config{
		User:                 "root",
		Net:                  "tcp",
		Addr:                 gitbase.Name,
		AllowNativePasswords: true,
		MaxAllowedPacket:     **32 * (2 << 10)**,
	}
`
Submitting a pull request to resolve.

I'll work on submitting something a little more interesting in the future.  :)

jamesnaftel:
Pull request:  [346](https://github.com/src-d/engine/pull/346)

PR #346, opened 2019-03-17
engine: sql: sets the MaxAllowedPacket to the desired 32M. Issue#345

jamesnaftel:
See issue https://github.com/src-d/engine/issues/345.

### DIFF
--- a/cmd/srcd-server/engine/sql.go
+++ b/cmd/srcd-server/engine/sql.go
@@ -36,7 +36,7 @@ func (s *Server) SQL(req *api.SQLRequest, stream api.Engine_SQLServer) error {
 		Net:                  "tcp",
 		Addr:                 gitbase.Name,
 		AllowNativePasswords: true,
-		MaxAllowedPacket:     32 * (2 << 10),
+		MaxAllowedPacket:     32 * (1 << 20),
 	}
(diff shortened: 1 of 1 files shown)


se7entyse7en:
LGTM, but most probably @smacker knows better as he recently worked on mysql client.

smacker:
@se7entyse7en my pull request changes how `engine-cli` works. But the server exposes gRPC methods that can be used by any other client.

cmd/srcd-server/engine/sql.go
```diff
@@ -36,7 +36,7 @@ func (s *Server) SQL(req *api.SQLRequest, stream api.Engine_SQLServer) error {
 		Net:                  "tcp",
 		Addr:                 gitbase.Name,
 		AllowNativePasswords: true,
-		MaxAllowedPacket:     32 * (2 << 10),
+		MaxAllowedPacket:     32 * (1 << 20),
```
carlosms: Thanks for noticing! I suggest to add a comment with the intended value, because the field name does not tell if it's in bytes, bits, MiB... 
```suggestion
		MaxAllowedPacket:     32 << 20, // 32 MiB
```

carlosms:
Closed in #387.

PR #350, opened 2019-03-19
Add integration tests for stop cmd

carlosms:
Part of #232.
Tests for `srcd stop`

### DIFF
--- a/cmd/srcd/cmd/stop_test.go
+++ b/cmd/srcd/cmd/stop_test.go
@@ -1,0 +1,88 @@
+// +build integration
+
+package cmd
+
+import (
+	"context"
+	"io/ioutil"
+	"log"
+	"os"
+	"testing"
+
+	cmdtest "github.com/src-d/engine/cmd/test-utils"
+	"github.com/src-d/engine/docker"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type StopTestSuite struct {
+	cmdtest.IntegrationSuite
+	testDir string
+}
+
+func TestStopTestSuite(t *testing.T) {
+	s := StopTestSuite{}
+	suite.Run(t, &s)
+}
+
+func (s *StopTestSuite) SetupTest() {
+	var err error
+	s.testDir, err = ioutil.TempDir("", "stop-test")
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (s *StopTestSuite) TearDownTest() {
+	s.RunStop(context.Background())
+	os.RemoveAll(s.testDir)
+}
+
+func (s *StopTestSuite) TestInitStop() {
+	require := s.Require()
+
+	_, err := s.RunInit(context.TODO(), s.testDir)
+	require.NoError(err)
+
+	_, err = s.RunSQL(context.TODO(), "SELECT 1")
+	require.NoError(err)
+
+	_, err = s.RunStop(context.TODO())
+	require.NoError(err)
+
+	s.AllStopped()
+}
+
+func (s *StopTestSuite) TestStopTwice() {
+	require := s.Require()
+
+	_, err := s.RunInit(context.TODO(), s.testDir)
+	require.NoError(err)
+
+	_, err = s.RunStop(context.TODO())
+	require.NoError(err)
+
+	_, err = s.RunStop(context.TODO())
+	require.NoError(err)
+}
+
+func (s *StopTestSuite) TestMissingContainers() {
+	require := s.Require()
+
+	_, err := s.RunInit(context.TODO(), s.testDir)
+	require.NoError(err)
+
+	// start gitbase and bblfsh
+	_, err = s.RunSQL(context.TODO(), "SELECT 1")
+	require.NoError(err)
+
+	// kill the daemon container
+	err = docker.RemoveContainer("srcd-cli-daemon")
+	require.NoError(err)
+
+	// run stop, the other containers should be stopped
+	_, err = s.RunStop(context.TODO())
+	require.NoError(err)
+
+	s.AllStopped()
+}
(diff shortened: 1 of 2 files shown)


cmd/srcd/cmd/stop_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (s *StopTestSuite) allStopped() {
+	s.T().Helper()
+	require := s.Require()
+
+	out, err := s.RunCommand(context.TODO(), "components", "list")
+	require.NoError(err, out.String())
+
+	expected := regexp.MustCompile(
```
dpordomingo: is the order guaranteed beforehand?

carlosms: Yes, the list of components is always [sorted here](https://github.com/src-d/engine/blob/9282f346450af6c6bc25884f1c7108b8ed17225f/components/components.go#L220).

cmd/srcd/cmd/stop_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// run stop, the other containers should be stopped
+	_, err = s.RunStop(context.TODO())
+	require.NoError(err)
+
+	s.allStopped()
+}
+
+func (s *StopTestSuite) allStopped() {
+	s.T().Helper()
```
dpordomingo: it seems to do the same than [IntegrationSuite::AllStopped()](https://github.com/src-d/engine/pull/351/files#diff-facfe3e96880ee046a919e28708f98e3R81)

cmd/srcd/cmd/stop_test.go
```diff
@@ -0,0 +1,108 @@
...
+	require.NoError(err)
+
+	// run stop, the other containers should be stopped
+	_, err = s.RunStop(context.TODO())
+	require.NoError(err)
+
+	s.allStopped()
+}
+
+func (s *StopTestSuite) allStopped() {
```
dpordomingo: it seems to do the same than [IntegrationSuite::AllStopped()](https://github.com/src-d/engine/pull/351/files#diff-facfe3e96880ee046a919e28708f98e3R81)

carlosms: Indeed, I should have mentioned that in #351.
I did this PR fist, then #351. My plan is to remove this `StopTestSuite.allStopped` and use `IntegrationSuite.AllStopped` in the rebase (either rebasing here, or in #351).

dpordomingo: imo both PRs could be merged in one; the review would be simpler and the common things would be easy to find ;)

se7entyse7en: I'm wondering whether we should use the docker sdk to check whether the components have been actually stopped. What I mean is that each test assumes that the stop command is working as they use that in their setup/teardown and the test of the stop command assumes that the components command works correctly. I agree that is practically a very rare case, but it could happen that the tests pass, but both stop and components are broken.

carlosms: I first wrote the tests using [`Component.IsRunning()`](https://github.com/src-d/engine/blob/master/components/components.go#L60). But then it looked to me like it was getting too coupled with the codebase components code. Specifically, it uses the codebase `image:version`, instead of the version provided by the `srcd` binary. For now we are testing the same code, but as much as possible I'd like to keep the tests able to run over any version of `srcd`.

I will take another look and see if I can come up with a clean way to use docker here.

carlosms: And if I submitted both PRs together we would probably be discussing the advantage of having them separated ¯\\_(ツ)_/¯

carlosms: Done in 0a683df

smacker: I was thinking more about
```
c, err := client.NewEnvClient()
containers := c.ContainerList()
assert("srcd-cli-bblfshd" not in containers)
```

carlosms: But then it gets a bit messy:

```go

func (s *StopTestSuite) allStopped() {
	s.T().Helper()
	require := s.Require()

	names := []string{
		"srcd-cli-bblfshd",
		"srcd-cli-bblfsh-web",
		"srcd-cli-daemon",
		"srcd-cli-gitbase-web",
		"srcd-cli-gitbase",
	}

	c, err := client.NewEnvClient()
	containers, err := c.ContainerList(context.Background(), types.ContainerListOptions{All: true})
	require.NoError(err)

	for _, container := range containers {
		for _, cName := range container.Names {
			for _, name := range names {
				cName = strings.TrimPrefix(cName, "/")
				require.NotEqual(name, cName, "Component %s should not be running", name)
			}
		}
	}
}
```

Our docker package is a wrapper that already deals with components list filter, multiple names for containers, the difference in name between `docker ps` and the docker api (`name` vs `/name`)...

Do you prefer the code in this comment?

smacker: No strong opinion on this. Originally I had a similar concern to yours:
> But then it looked to me like it was getting too coupled with the codebase components code

this time about docker package. But I re-read the code and it looks general enough to use in the tests 👍

carlosms: I'd prefer to leave it using the docker package. Maybe tomorrow we update the lib and the names come without a '/'  prefix, or ContainerList requires some new option to list things properly and we forget to update the tests.

se7entyse7en: Yup using the raw docker sdk seems messy and as you pointed out by using our docker package we need to change only that in case we update the docker sdk lib.

carlosms:
Done, I moved the code to `cmd/test-utils/IntegrationSuite::AllStopped()` and rebased.

Issue #347, opened 2019-03-19
Refactor integration test code


carlosms:
Once every PR for #232  is merged we should consider some improvements.

Some examples:
- Tests don't need to be `package cmd`, they don't use any code from `cmd` and should be at least `cmd_test`.
- Since it's integration testing, and not really testing the code in `cmd/srcd/cmd/`, should we move the files to other folder?
- Maybe it's worth to make a common `SetupTest` with the tmpdir creation.
- Improve `IntegrationSuite.CommandContext`, `IntegrationSuite.RunCommand`, `ParseTestSuite.runCommandStdout`...
- Some commands have a wrapper that don't offer any extra functionality, like `IntegrationSuite.RunStop`. For some other commands, we call `RunCommand(context.TODO(), "prune")`. We should either remove the wrappers, or add wrappers for all commands. I prefer to remove them, except when they provide some kind of syntactic sugar or added benefit.
- See if it makes sense to use https://godoc.org/gotest.tools/icmd

dpordomingo:
some nice-to-have, imo

- [ ] https://github.com/src-d/engine/pull/324#discussion_r265705163 separate `stderr` and `stdout`: 
_When running commands in tests, &mdash;like done by [IntegrationSuite::RunCommand](https://github.com/src-d/engine/blob/b72f1fe345a6f035457dfcf4658b92cc65179f45/cmd/test-utils/common.go#L42)&mdash; `stderr` and `stdout` should be separated and assert that nothing is written in `stderr` if the command produced no errors, and the opposite in such cases._
- [ ] https://github.com/src-d/engine/pull/315#discussion_r264764563 more standard way to assert the _Changes done by Engine in its components: 
components outputs could be extracted in a standard way, being able to assert the status of every component in a more readable way (example by https://play.golang.org/p/bC_LNM02Hi-)_

PR #351, opened 2019-03-19
Add integration tests for prune cmd

carlosms:
Part of #232.
Tests for `srcd prune`.

This test suite does not test the `--with-images` flag.
If the tests are run with `make test-integration` the daemon image will be the one build locally (e.g. srcd/cli-daemon:dev-b72f1fe), and deleting this image would make all the other tests fail.

### DIFF
--- a/cmd/srcd/cmd/prune.go
+++ b/cmd/srcd/cmd/prune.go
@@ -16,7 +16,11 @@ var pruneCmd = &cobra.Command{
 			return humanizef(err, "could not prune components")
 		}
 
-		return humanizef(daemon.CleanUp(), "could not clean up")
+		if err := daemon.CleanUp(); err != nil {
+			return humanizef(err, "could not clean up")
+		}
+
+		return nil
 	},
 }
 
--- a/cmd/srcd/cmd/prune_test.go
+++ b/cmd/srcd/cmd/prune_test.go
@@ -1,0 +1,146 @@
+// +build integration
+
+package cmd
+
+import (
+	"context"
+	"io/ioutil"
+	"log"
+	"os"
+	"sort"
+	"testing"
+
+	cmdtest "github.com/src-d/engine/cmd/test-utils"
+	"github.com/src-d/engine/docker"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/suite"
+)
+
+// This test suite does not test the `--with-images` flag.
+// If the tests are run with `make test-integration` the daemon image will be
+// the one build locally (e.g. srcd/cli-daemon:dev-b72f1fe), and deleting
+// this image would make all the other tests fail.
+
+type PruneTestSuite struct {
+	cmdtest.IntegrationSuite
+	testDir string
+}
+
+func TestPruneTestSuite(t *testing.T) {
+	s := PruneTestSuite{}
+	suite.Run(t, &s)
+}
+
+func (s *PruneTestSuite) SetupTest() {
+	var err error
+	s.testDir, err = ioutil.TempDir("", "prune-test")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// The tests require a clean starting point. Relaying on the prune command
+	// itself to provide the clean starting point is not ideal, but it's the
+	// best option for now
+	out, err := s.RunCommand(context.TODO(), "prune")
+	s.Require().NoError(err, out.String())
+}
+
+func (s *PruneTestSuite) TearDownTest() {
+	s.RunStop(context.Background())
+	os.RemoveAll(s.testDir)
+}
+
+func (s *PruneTestSuite) TestRunningContainers() {
+	require := s.Require()
+
+	// Get the list of volumes and networks before calling init
+	prevVols, err := docker.ListVolumes(context.Background())
+	require.NoError(err)
+
+	prevNets, err := listNetworks()
+	require.NoError(err)
+
+	_, err = s.RunInit(context.TODO(), s.testDir)
+	require.NoError(err)
+
+	_, err = s.RunSQL(context.TODO(), "SELECT 1")
+	require.NoError(err)
+
+	out, err := s.RunCommand(context.TODO(), "prune")
+	require.NoError(err, out.String())
+
+	// Test containers were deleted
+	s.AllStopped()
+
+	// Test volumes with name srcd-cli-* were deleted.
+	// This does not actually test much because Engine does not create volumes
+	// since v0.7.0.
+	// The logic used in prune to delete named volumes is looking for the
+	// srcd-cli- prefix in the name, so that's what we check here.
+	vols, err := docker.ListVolumes(context.Background())
+	require.NoError(err)
+
+	for _, vol := range vols {
+		require.NotContainsf(vol.Name, "srcd-cli-", "Volume %s was not deleted", vol.Name)
+	}
+
+	// Test anonymous volumes were deleted
+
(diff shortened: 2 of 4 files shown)


cmd/srcd/cmd/prune_test.go
```diff
@@ -0,0 +1,65 @@
...
+func (s *PruneTestSuite) TestRunningContainers() {
+	require := s.Require()
+
+	_, err := s.RunInit(context.TODO(), s.testDir)
+	require.NoError(err)
+
+	_, err = s.RunSQL(context.TODO(), "SELECT 1")
+	require.NoError(err)
+
+	_, err = s.RunCommand(context.TODO(), "prune")
```
dpordomingo: The integration tests for stop command use [`IntegrationSuite:RunStop()`](https://github.com/src-d/engine/pull/350/files#diff-6fd0a789630fde34454c1121b58bdf57R51); for consistency, this test could use a new `IntegrationSuite:RunPrune()`

cmd/test-utils/common.go
```diff
@@ -76,6 +78,25 @@ func (s *IntegrationSuite) RunStop(ctx context.Context) (*bytes.Buffer, error) {
 	return s.RunCommand(ctx, "stop")
 }
 
+func (s *IntegrationSuite) AllStopped() {
```
dpordomingo: I don't see the difference with [`StopTestSuite::allStopped()`](https://github.com/src-d/engine/pull/350/files#diff-6fd0a789630fde34454c1121b58bdf57R91)

cmd/srcd/cmd/prune_test.go
```diff
@@ -0,0 +1,65 @@
...
+	require.NoError(err)
+
+	s.AllStopped()
```
smacker: for prune we should test not only that containers were stopped but that other stuff was removed (network, volumes)

smacker:
@dpordomingo https://github.com/src-d/engine/issues/352

cmd/test-utils/common.go
carlosms: Answered in https://github.com/src-d/engine/pull/350#discussion_r266928822.

cmd/srcd/cmd/prune_test.go
carlosms: Honestly, I dislike methods like this
```go
func (s *IntegrationSuite) RunStop(ctx context.Context) (*bytes.Buffer, error) {
	return s.RunCommand(ctx, "stop")
}
```

And I would prefer to not add any new ones in this PR.

This is something we can discuss for #347, to either remove them all or add missing ones (like `RunPrune`).

dpordomingo: I think it's better to look for consistency in all iterations, bec then it's easier to clean later, but fair enough; I added it to my wishlist @ https://github.com/src-d/engine/issues/347#issuecomment-474337757

carlosms:
I've pushed new checks for the volumes and network in a782fcb.

> also, there is no way to test --with-images?
> maybe in a different Travis stage?

Maybe... or it could be the same travis stage, but a separate command we can run after `make test-integration`. Like `WITH_IMAGE_PRUNE=true make test-integration-no-build`. This would allow to skip the prune test when you run it locally. Otherwise it can be inconvenient to have the tests removing the images every time you run them on your machine.

What do you think?

cmd/srcd/cmd/prune_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// Test volumes were deleted.
+	// This does not actually test much because Engine does not create volumes
+	// since v0.7.0.
+	// The logic used in prune to delete volumes is looking for the
+	// srcd-cli- prefix in the name, so that's what we check here.
+	vols, err := docker.ListVolumes(context.Background())
+	require.NoError(err)
+
+	for _, vol := range vols {
+		require.NotContainsf(vol.Name, "srcd-cli-", "Volume %s was not deleted", vol.Name)
```
dpordomingo: I don't know the docker interface, so it might have no sense my question, but... Why not ensuring that there is no `srcd-cli-*` volumes in the returned list of volumes?

cmd/srcd/cmd/prune_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Test networks were deleted.
+	nets, err := listNetworks()
+	require.NoError(err)
+
+	for _, net := range nets {
+		require.NotEqualf(net.Name, docker.NetworkName, "Network %s was not deleted", net.Name)
```
dpordomingo: would not be better to ensure that there is no `srcd-cli-network` left after pruning?

carlosms:
> Also, I tried srcd prune after init, and I still see volumes there.

@dpordomingo good catch, I have created https://github.com/src-d/engine/issues/371 to fix that. I also added 11a2992 to test any dangling volume or network, besides checking the ones we know we create with proper names.

cmd/srcd/cmd/prune_test.go
carlosms: I don't understand your question. That is what the code is doing, going over the list of of volumes and checking the name is not `srcd-cli-*`.

cmd/srcd/cmd/prune_test.go
carlosms: same as the other comment, I don't understand what you mean. That is what the code is doing, checking `srcd-cli-network` is not in the list of networks after pruning.

cmd/test-utils/common.go
carlosms: I rebased the branch, now it's using the common code from #350.

cmd/srcd/cmd/prune_test.go
dpordomingo: I'm absolute stupid :100: 
just disregard this comment ;)

cmd/srcd/cmd/prune_test.go
dpordomingo: same shame :1st_place_medal:

cmd/srcd/cmd/prune.go
```diff
@@ -16,7 +16,11 @@ var pruneCmd = &cobra.Command{
 			return humanizef(err, "could not prune components")
 		}
 
-		return humanizef(daemon.CleanUp(), "could not clean up")
+		if err := daemon.CleanUp(); err != nil {
```
smacker: I would better add `if` inside `humanizef` function. It will make client code cleaner and we will avoid similar errors in other places.

carlosms:
> since #355, `prune` command also deletes gitbase indexes, so I think this integration tests should also assert it is properly done.

We already have #373, when we merge it there will not be gitbase indexes in `~/.srcd`. With 11a2992 we should be able to detect if any extra volume is left-over.
Would you agree to merge this PR as it is, and open a new issue for new tests for `--with-images`?

cmd/srcd/cmd/prune.go
carlosms: I have pushed that `if` in ddecbcb.

I don't think we should change the caller style in this PR, it looks like all the rest of the code does
```go
err := call()
if err != nil {
  return humanizef(err, "msg")
}
```
instead of 
```go
return humanizef(call(), "msg")
```

If you think we should change it I'd rather move that discussion to slack or a new issue. And then change all the calls in the same PR.

smacker: I'm okay with both snippets. The important thing: using any of them doesn't panic anymore :)

smacker:
Imo we should merge it now and implement the rest in separate issues.

carlosms:
New issue for `--with-images` in #374.
Branch rebased and squashed, merging.

Issue #300, opened 2019-03-06
Let the user know under which port is each service


dpordomingo:
Since services can be accessed from [Clients and Connectors](https://docs.sourced.tech/engine/#clients-and-connectors), the container ports are publicly exposed.

It would be great if they could be listed somewhere, for example with `srcd components list`
```shell
$ srcd components list
IMAGE                           INSTALLED  RUNNING  PORT  CONTAINER NAME
bblfsh/bblfshd:v2.11.8-drivers  yes        yes      9432  srcd-cli-bblfshd
bblfsh/web:v0.9.0               no         no             srcd-cli-bblfsh-web
srcd/cli-daemon:v0.11.0-rc.1    yes        yes      4242  srcd-cli-daemon
srcd/gitbase-web:v0.6.0         no         no             srcd-cli-gitbase-web
srcd/gitbase:v0.19.0-rc1        yes        yes      3306  srcd-cli-gitbase
```

PR #354, opened 2019-03-20
show public ports of components in components list

smacker:
fix: #300

Integration test checks that the correct port is displayed in `TestListInit`.
Please let me know if you think I should remove it from there and add another test case especially for the ports.

### DIFF
--- a/components/components.go
+++ b/components/components.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/docker/api/types"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/src-d/engine/docker"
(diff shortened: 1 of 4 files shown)


components/components.go
```diff
@@ -61,6 +62,19 @@ func (c *Component) IsRunning() (bool, error) {
 	return docker.IsRunning(c.Name, c.ImageWithVersion())
 }
 
+// GetPorts returns component ports of the component if there is any
+func (c *Component) GetPorts() ([]types.Port, error) {
```
carlosms: For consistency, instead of dealing with docker api types in the cli, maybe we could do `GetPublicPorts() ([]uint16, error)` and simplify `publicPortsFmt`.

smacker: it was my original idea and actually I even implement it like that first. But then I figured we might need a private port in the future.

for example to connect from `mysql-client` to gitbase inside the docker you need to know private port. Currently, it just matches default one but we might need something similar in the future for other components.

To solve this issue we can actually hard-code internal port because we pin all the images. But if we are going to implement something like https://github.com/src-d/backlog/issues/1364 we will need this method.

carlosms: I see. Should we do a type alias for `types.Port`? In our docker package we do 
```go
type Container = types.Container
type Volume = types.Volume
```

But in this case I don't know if we should do `docker.Port` (consistent with other docker types aliases, but the method  is in `components` package), a `components.Port`, or just leave it as it is...

smacker:
rebased on master to fix flaky integration test.
check for `go mod vendor` is still flaky. If it fails, please just ignore it for this PR.

components/components.go
smacker: No opinion on it. I'm good with adding the alias as well as keeping using `docker/types`.

//cc @se7entyse7en wdyt?

se7entyse7en: I also prefer the alias and I'd put it in the `docker` package. IMO is more suitable there instead of adding it in the `components` package.

smacker:
alias added

Issue #352, opened 2019-03-19
srcd prune should remove gitbase indexes too


smacker:
Idea from https://github.com/src-d/engine/pull/351#pullrequestreview-216177409

Makes a lot of sense to me.

dpordomingo:
Does source{d} Engine use temporary data in the host machine?
some other thing in `~/.gitbase` that should be deleted?

carlosms:
I think we only have the `~/.srcd/gitbase` dir.

PR #355, opened 2019-03-20
prune: remove gitbase indexes

smacker:
fix: #352 

there is no test for it because prune tests pr isn't merged to the master yet

### DIFF
--- a/cmd/srcd/daemon/daemon.go
+++ b/cmd/srcd/daemon/daemon.go
@@ -165,6 +165,8 @@ func createDaemon(workdir string) docker.StartFunc {
 		if err != nil {
 			return err
 		}
+		// we run the command inside docker so slashes must be always converted to unix-style
+		datadir = filepath.ToSlash(datadir)
 
 		if err := setupDataDirectory(workdir, datadir); err != nil {
 			return err
@@ -217,5 +219,5 @@ func datadir() (string, error) {
 		return "", errors.Wrap(err, "unable to get home dir")
 	}
 
-	return filepath.ToSlash(filepath.Join(homedir, ".srcd")), nil
+	return filepath.Join(homedir, ".srcd"), nil
 }
--- a/cmd/srcd/daemon/daemon.go
+++ b/cmd/srcd/daemon/daemon.go
@@ -65,6 +65,18 @@ func Kill() error {
 	return nil
 }
 
+// CleanUp removes all resources created by daemon on host
+func CleanUp() error {
+	datadir, err := datadir()
+	if err != nil {
+		return err
+	}
+
+	gitbaseIndexDir := filepath.Join(datadir, "gitbase")
+
+	return os.RemoveAll(gitbaseIndexDir)
+}
+
 // Client will return a new EngineClient to interact with the daemon. If the
 // daemon is not started already, it will start it at the working directory.
 func Client() (api.EngineClient, error) {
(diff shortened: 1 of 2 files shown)


cmd/srcd/daemon/daemon.go
```diff
@@ -66,6 +66,18 @@ func Kill() error {
...
 }
 
+// CleanUp removes all resources created by daemon on host
+func CleanUp() error {
+	datadir, err := datadir()
+	if err != nil {
+		return err
+	}
+
+	gitbaseIndexDir := filepath.Join(datadir, "gitbase")
```
carlosms: I wonder if this will work on windows. `datadir()` uses `/` -`filepath.ToSlash(filepath.Join(homedir, ".srcd"))`-, then here we join with the normal windows separator `\`.

smacker: true. We should do `filepath.ToSlash` only inside daemon even if it works. Thanks!

cmd/srcd/daemon/daemon.go
```diff
@@ -166,6 +166,8 @@ func createDaemon(workdir string) docker.StartFunc {
 		if err != nil {
 			return err
 		}
+		// we run the command inside docker so slashes must be always converted to unix-style
```
carlosms: Is this right? this is run by the client, on the host, not inside a container.

smacker: we run the command on client but the value is used inside container: https://github.com/src-d/engine/pull/355/files#diff-f65e86a43ab65d559e1ce8134b4aa922R197
or am I wrong?

cmd/srcd/daemon/daemon.go
se7entyse7en: True that this should be tested on windows.

> `datadir()` uses `/`

Are we sure about this? Isn't `homedir.Dir()` inside `datadir()` returning a path separated by `\`?

@smacker what do you mean with "even if it works"?

se7entyse7en: Ok, I got it now. The code this thread was pointing was the old one 😅

cmd/srcd/daemon/daemon.go
carlosms: ok yes, I understood "the command" as the command calling `createDaemon()`

smacker:
rebased on master, merge after CI pass

cmd/srcd/daemon/daemon.go
```diff
@@ -65,6 +65,18 @@ func Kill() error {
...
 }
 
+// CleanUp removes all resources created by daemon on host
+func CleanUp() error {
+	datadir, err := datadir()
+	if err != nil {
+		return err
+	}
+
+	gitbaseIndexDir := filepath.Join(datadir, "gitbase")
```
dpordomingo: I assume we're only storing there only the default config file, and `gitbase` indexes.
The problem with current implementation is that if I understood it properly, we're not ensuring now that `prune` will delete everything that Engine could create in host machine, but only the gitbase indexes (what is ok in terms of the issue, but not in terms of `prune` command)

I'd suggest to separate `config` and `data` dirs, so we could delete the whole `data` dir with `prune`, also all new things that could be created by Engine at some point.

If you agree with this suggestion, let me know and I'll open an issue with it.

PR #363, opened 2019-03-22
Gitbase multi version testing

### LINKED ISSUE #336: Test for gitbase index data, and gitbase updates
Right now engine does not handle what happens when we upgrade gitbase, and there is existing index data in `~/.srcd`.
This issue is to define a test that can detect such incompatibility.
It could be done installing a previous `srcd` version, creating indexes, then using the current `srcd` we are testing to start gitbase and use those indexes.
Or we could also have some fixtures for this data.

se7entyse7en:
Closes #336.

This PR adds integration testing that tests the compatibility of gitbase across different version of engine:
```
make test-integration
```

Two versions are considered that are named `PREV` and `NEXT`. By default `PREV` is the latest non rc version and `NEXT` is `HEAD`. For testing it is also possible to pass this as env vars in order to test two arbitrary version of engine:
```
PREV_VERSION=v0.10.0 NEXT_VERSION=v0.11.0 make test-integration
```


se7entyse7en:
I was thinking about a couple of things. Given that it is possible to test two given version of engine for retro-compatibility, and that this makes sense only for this kind of tests, maybe we could add another build flag such as `integration-retro-comp` and run them separately from the "simple" integration tests. So that we have:
```
make test-integration
```
and
```
make test-integration-retro-comp
```

WDYT?

se7entyse7en:
BTW by running this:
```
PREV_VERSION=v0.10.0 NEXT_VERSION=v0.11.0 make test-integration
```
I was able to detect the backward incompatibility. 🎉 
As other examples, both this:
```
PREV_VERSION=v0.10.0 NEXT_VERSION=v0.11.0 make test-integration
```
and this:
```
PREV_VERSION=v0.11.0 make test-integration
```
succeeded.

se7entyse7en:
@smacker I didn't add the `EXPLAIN TREE` thing as I think that the current test is enough, WDYT?

se7entyse7en:
mhmmm for some reason it didn't execute the integration tests on travs... I'll investigate.

smacker:
> I didn't add the EXPLAIN TREE thing as I think that the current test is enough, WDYT?

I would better add it to make sure that the index is actually used by gitbase. Otherwise, they may show index in "SHOW INDEXES" but because it's outdated/calculated for a different directory they skip it when executing the query. But it's not like VERY important.

Makefile
```diff
@@ -18,6 +18,52 @@ $(MAKEFILE):
...
+	rm -rf $(CURRENT_BUILD_PATH)
+	$(GOCLEAN) .
+integration-prepare-build-current:
+	git checkout -q $(CURRENT_VERSION)
+	$(eval BUILD_PATH = $(CURRENT_BUILD_PATH))
+integration-build-current: integration-clean-current integration-prepare-build-current build docker-build
+	git checkout -q $(CURRENT_VERSION)
+	echo "Built version $(CURRENT_VERSION)"
+
+integration-build-all:
```
smacker: have you thought about using pre-built releases from github instead of building them locally?
I'm not sure what is better.
Current approach is more flexible. But using pre-built releases, in theory, are better reproducible. (something can change over time and the local build won't be exact the same as the one from github)

se7entyse7en: > Current approach is more flexible.

Actually, I think that it doesn't have much sense to test a prev version that is not an actual release available on GH (I don't think that we're going to test a specific unreleased version as `PREV`).  And as you said we're actually interested in the release available on GH. Even if the build artifact should be exactly the same, downloading it should also speed-up testing and aslo won't require building the docker image for the daemon.

se7entyse7en: Given that this PR is already big, do you mind approaching this separately?

cmd/test-utils/common.go
```diff
@@ -18,31 +18,27 @@ import (
...
 
-func (s *IntegrationSuite) RunCommand(ctx context.Context, cmd string, args ...string) (*bytes.Buffer, error) {
-	var out bytes.Buffer
-
-	command := s.CommandContext(ctx, cmd, args...)
-	command.Stdout = &out
-	command.Stderr = &out
+type IntegrationSuite struct {
+	suite.Suite
+	Cmd *Commander
```
carlosms: I think this could have been embedded to avoid many changes to the existing tests
```go
type IntegrationSuite struct {
	suite.Suite
	Commander
}
```

se7entyse7en: Woa! For some reason I was thinking that only one struct can be embedded! 😂
Yeah that should minimize the changes!

cmd/srcd/cmd/gitbase_back_comp_test.go
```diff
@@ -0,0 +1,222 @@
...
+	s.Cmd.RunStop(context.Background())
+	s.PrevCmd.RunStop(context.Background())
+	os.RemoveAll(s.testDir)
+	os.RemoveAll(s.testDataDir)
+}
+
+var newLineFormatter = regexp.MustCompile(`(\r\n|\r|\n)`)
+var reposExtractors = regexp.MustCompile(`(?:\|\s+([\w-]+)\s+\|\n)`)
+
+func (s *GitbaseBackCompTestSuite) getBufString(buf *bytes.Buffer) string {
```
carlosms: this is the same as `normalizeNewLine` in `common.go`?

carlosms:
I have a similar comment to @smacker. I didn't test the current approach extensively, but my first impression is that it limits the usefulness of the integration tests in the local dev machine.
Using the current git to checkout older versions has these problems:
- It does not return to the branch I was working on. Instead, it checks out the last commit.
- If I have any local changes I cannot run the integration tests.

Did you check if we can use https://github.com/src-d/regression-core to download or build the binaries?

se7entyse7en:
> It does not return to the branch I was working on. Instead, it checks out the last commit.
true, this could be fixed easily I think.
> If I have any local changes I cannot run the integration tests.
true, I think that this could be solved by stashing/unstashing.

But both things are actually not needed if we decide to test releases versions (which makes a lot of sense actually).

> Did you check if we can use https://github.com/src-d/regression-core to download or build the binaries?

I forgot about it 😕. I'm gonna take a look into it 👍.

se7entyse7en:
This is currently on hold for [this](https://github.com/src-d/regression-core/issues/8).

smacker:
> Given that this PR is already big, do you mind approaching this separately?

if we go with src-d/regression-core approach it will make sense to follow the same workflow as other regression tests (a separate binary that runs on jenkins).

Another thing if we embed struct correctly PR won't be that big & the code will be very localized without affecting anything else. So I don't think there should be a rush with merging.

se7entyse7en:
> So I don't think there should be a rush with merging.

👍 

> if we go with src-d/regression-core approach it will make sense to follow the same workflow as other regression tests (a separate binary that runs on jenkins).

Yup, I was thinking the same, we should discuss whether it makes more sense to do as you said.

.travis.yml
```diff
@@ -29,10 +29,12 @@ jobs:
       script:
         - make packages
         - make test-coverage
-
     - name: 'Integration Tests Linux'
       script:
         - make test-integration
+    - name: 'Regression Tests Linux'
```
smacker: let's run in only on master in background. Or only by schedule if it's possible.

cmd/srcd/cmd/gitbase_back_comp_test.go
```diff
@@ -0,0 +1,312 @@
...
+| repositories |          1 | repo_idx |            1 | repositories.repository_id | NULL      |           0 |        0 | NULL   |      | pilosa     |         |               | YES     | NULL       |
++--------------+------------+----------+--------------+----------------------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+------------+
+`
+	require.NoError(cmdtest.AreSQLOutputEqual(buf.String(), expected))
+
+	// [previous version] srcd sql "select * from repositories"
+	buf, err = s.runSQL(s.PrevCmd, "select * from repositories")
+	require.NoError(err)
+
+	expected = `+---------------+
```
dpordomingo: Is this not the same than https://github.com/src-d/engine/pull/363/files#diff-0a916376db6c11d729358f6303375940R206 ?
Why is it needed? could it be different after the index creation?
Is it needed this test for the purpose of the backwardcompatibility test?

.travis.yml
se7entyse7en: As discussed on slack, I'm gonna remove it 👍.

cmd/srcd/cmd/gitbase_back_comp_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// [current version] srcd init s.testDir
+	_, err = s.runInit(s.CurrCmd, s.testDir)
+	require.NoError(err)
+
+	// [current version] srcd sql "select * from repositories"
+	buf, err = s.runSQL(s.CurrCmd, "select * from repositories")
+	require.NoError(err)
+
+	expected = `+---------------+
```
dpordomingo: if the purpose is to ensure that indexes could be used, why not checking it explicitly?
```sql
describe format=tree select repository_id from repositories where repository_id = 'repo-0'
```
should answer:
```
+-----------------------------------------------------+
|                      PLAN                           |
+-----------------------------------------------------+
| Exchange(parallelism=4)                             |
|  └─ Table(repositories)                             |
|      ├─ Column(repository_id, TEXT, nullable=false) |
|      ├─ Filters                                     |
|      │   └─ repositories.repository_id = "repo-0"   |
|      └─ Indexes                                     |
|          └─ repo_idx                                |
+-----------------------------------------------------+
```

se7entyse7en:
> also after the long discussion, and the rebase push force I'm not sure about the state of this PR

Yup sorry for the force push, but there were so many changes that it would make the following rebase much more difficult.

cmd/srcd/cmd/gitbase_back_comp_test.go
se7entyse7en: addressed

cmd/srcd/cmd/gitbase_back_comp_test.go
```diff
@@ -0,0 +1,312 @@
...
+| TABLE | NON UNIQUE | KEY NAME | SEQ IN INDEX | COLUMN NAME | COLLATION | CARDINALITY | SUB PART | PACKED | NULL | INDEX TYPE | COMMENT | INDEX COMMENT | VISIBLE | EXPRESSION |
++-------+------------+----------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+------------+
++-------+------------+----------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+------------+
+`
+	require.NoError(cmdtest.AreSQLOutputEqual(buf.String(), expected))
+
+	// [previous version] srcd sql "CREATE INDEXES"
+	_, err = s.runSQL(s.PrevCmd, "CREATE INDEX repo_idx ON repositories USING pilosa (repository_id)")
+	require.NoError(err)
+	// wait a bit for index to be ready
```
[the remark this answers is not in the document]
se7entyse7en: In this simple case (creating an index for 3 empty repos) I actually think that it's ok, but I totally agree that having a more formal way is better. If you think this to be a blocker I'm gonna take a look at it, otherwise I'd keep this as-is and eventually open an issue. I don't know if we create some index also somewhere else.

cmd/srcd/cmd/gitbase_back_comp_test.go
se7entyse7en: mhmmm I think that at some point during the development I thought of checking the indexes explicitly. Given that also @smacker pointed out checking the index explicitly as you, I'm gonna address it 👍.

cmd/srcd/cmd/gitbase_back_comp_test.go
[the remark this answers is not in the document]
dpordomingo: maybe not a blocker for this PR, but I'd open an issue to do it at some point because I see the value of doing these tests with more real cases, where indexes will take more than 1sec to be ready :smile:

cmd/srcd/cmd/gitbase_back_comp_test.go
se7entyse7en: done

cmd/srcd/cmd/gitbase_back_comp_test.go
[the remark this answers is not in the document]
se7entyse7en: I've created [this issue](https://github.com/src-d/engine/issues/378).

se7entyse7en:
> `setupWd` is a dirty hack but I think it's okay to keep it for now

Yeah, I know 😞. I've realized later that `go test` sets the working dir to the package being tested, because on the other hand `dlv test` doesn't. We could eventually open an issue on regression-core to avoid assuming that it runs from the root of the project.

carlosms:
What is the status of this PR? `vendor` is not updated. Is this because you are waiting on https://github.com/src-d/regression-core/issues/8?

se7entyse7en:
> What is the status of this PR? vendor is not updated. Is this because you are waiting on src-d/regression-core#8?

Yup, I was able to run it locally by patching it.

smacker:
pinged guys in #dev-retrieval.

se7entyse7en:
> pinged guys in #dev-retrieval.

Thank you. 

@jfontan have just merged the PR in regression-core, I rebased and force-pushed. Gonna merge once all green.

carlosms:
Can you please wait a little bit? I'd like to review this one.

se7entyse7en:
> Can you please wait a little bit? I'd like to review this one.

Sure! It's not urgent, so just take your time ;)

carlosms:
This test is failing on master, and also locally for me:
```
$ go test -v -parallel 1 -count 1 -ldflags "-X main.version=dev-91f80a0-dirty -X main.build=04-03-2019_16_28_00 -X main.commit=91f80a0" -tags=integration github.com/src-d/engine/cmd/srcd/cmd/ -run TestPruneTestSuite
=== RUN   TestPruneTestSuite
=== RUN   TestPruneTestSuite/TestRunningContainers
=== RUN   TestPruneTestSuite/TestStoppedContainers
--- FAIL: TestPruneTestSuite (0.00s)
    --- FAIL: TestPruneTestSuite/TestRunningContainers (0.00s)
        suite.go:61: test panicked: runtime error: invalid memory address or nil pointer dereference
    --- FAIL: TestPruneTestSuite/TestStoppedContainers (0.00s)
        suite.go:61: test panicked: runtime error: invalid memory address or nil pointer dereference
FAIL
FAIL	github.com/src-d/engine/cmd/srcd/cmd	0.005s
```

carlosms:
Is `REG_TOKEN` really needed? Why is it not optional? I think most github API actions can be done without auth, with something like a limit of 60 requests/hour. But it should be fine to run it locally, specially if we have a cache of downloaded binaries.

And I think the env var `REG_TOKEN` should be named `GITHUB_TOKEN`. It is the name we are already using in https://github.com/src-d/ci/tree/v1, and it's already configured in travis.

Since we are not adding this yet to travis, I also miss some docs on how to use it.
`PREV_ENGINE_VERSION` is buried in the code. We should document how to use `make test-regression` somewhere, like https://github.com/src-d/engine/blob/master/CONTRIBUTING.md. Or at least `test-regression` could have an `echo` with an example, I think that could also be enough for now.

When I've tried to run it locally it has failed with this error:

```
$ REG_TOKEN="xxxx" PREV_ENGINE_VERSION=v0.10.0 make test-regression
go test -v -parallel 1 -count 1 -ldflags "-X main.version=dev-91f80a0 -X main.build=04-03-2019_17_06_13 -X main.commit=91f80a0" -tags=regression github.com/src-d/engine/cmd/srcd/cmd/
=== RUN   TestGitbaseBackCompTestSuite
=== RUN   TestGitbaseBackCompTestSuite/TestRetroCompatibleIndexes
--- FAIL: TestGitbaseBackCompTestSuite (82.73s)
    --- FAIL: TestGitbaseBackCompTestSuite/TestRetroCompatibleIndexes (71.77s)
        require.go:794: 
            	Error Trace:	gitbase_back_comp_test.go:205
            	Error:      	Received unexpected error:
            	            	signal: killed
            	Test:       	TestGitbaseBackCompTestSuite/TestRetroCompatibleIndexes
FAIL
FAIL	github.com/src-d/engine/cmd/srcd/cmd	82.749s
Makefile:50: recipe for target 'test-regression' failed
make: *** [test-regression] Error 1error:
```
It only works if I manually pull all the required images. I guess bblfsd, which is the bigger one, takes a lot of time to download and there is a timeout somewhere?

se7entyse7en:
> This test is failing on master, and also locally for me

Gonna take a look into it.

> Is `REG_TOKEN` really needed? Why is it not optional? I think most github API actions can be done without auth, with something like a limit of 60 requests/hour.

Gonna take a look at `regression-core` about that `token`.

> But it should be fine to run it locally, specially if we have a cache of downloaded binaries.

Yup, `regression-core` provides a cache mechanism.

> Since we are not adding this yet to travis, I also miss some docs on how to use it.

You're right, I'm gonna add it to `CONTRIBUTING.md` and also add an `echo`.

> It only works if I manually pull all the required images. I guess bblfsd, which is the bigger one, takes a lot of time to download and there is a timeout somewhere?

I don't know actually, it never happened to me locally. Gonna investigate.

se7entyse7en:
@carlosms 

1. the test that was failing on `master` was because that test didn't exist before the rebase and in this PR there were some changes in the embedded `cmdtest.IntegrationSuite` `struct`. Fixed [here](https://github.com/src-d/engine/pull/363/commits/b96e4b96493f075552ac306234f2cae1a850cd80).
2. the regression test was failing for a timeout that was set to `1 * time.Minute` 🤦‍♂️ which is surely not enough in case of missing images. Most probably I never run it by pruning the images first. Fixed [here](https://github.com/src-d/engine/pull/363/commits/060aa15306a3b10e1ad7c174b9816de584eb15f6).
3. about `REG_TOKEN` you're right, it's not required, thanks for noticing. Renamed it into `GITHUB_TOKEN` [here](https://github.com/src-d/engine/pull/363/commits/f75848d219c18b14a8ee72f0abbfa0a609f698d8).
4. I noticed that `CONTRIBUTING.md` currently completely lacks explanation about how to run tests in general, so for now I've just added a usage explanation for the Makefile target. Maybe we could add an issue to improve our `CONTRIBUTING.md`. Fixed [here](https://github.com/src-d/engine/pull/363/commits/928a230d6590deb30413eeb99f4c6a07fa7fcc74).

dpordomingo:
Agree about docs. I opened https://github.com/src-d/engine/issues/408 to enhance `CONTRIBUTING.md` and some other issues I think we have :D

carlosms:
I don't think we should skip the test if the gitbase image is the same version.
Take for instance Engine 0.11 and 0.12. We changed the index data storage from a dir to volumes, and we know it's a breaking change that will make you loose the previous data when you upgrade. But the test right now does not detect that:
```
$ PREV_ENGINE_VERSION=v0.11.0 make test-regression

Usage: `PREV_ENGINE_VERSION=<first engine version to compare (default: 'latest')> CURR_ENGINE_VERSION=<second engine version to compare (default: 'local:HEAD')> make test-regression`
Examples:
- `make test-regression`                                                          # tests that latest version is forward-compatible with current (HEAD) version
- `PREV_ENGINE_VERSION=v0.10.0 make test-regression`                              # tests that v0.10.0 version is forward-compatible with current (HEAD) version
- `PREV_ENGINE_VERSION=v0.10.0 CURR_ENGINE_VERSION=v0.11.0 make test-regression`  # tests that v0.10.0 version is forward-compatible with v0.11.0 version

go test -v -parallel 1 -count 1 -ldflags "-X main.version=dev-928a230-dirty -X main.build=04-04-2019_16_56_05 -X main.commit=928a230" -tags=regression github.com/src-d/engine/cmd/srcd/cmd/
=== RUN   TestGitbaseBackCompTestSuite
--- SKIP: TestGitbaseBackCompTestSuite (19.48s)
    gitbase_back_comp_test.go:98: Gitbase didn't change
PASS
ok  	github.com/src-d/engine/cmd/srcd/cmd	19.485s
```

se7entyse7en:
@carlosms You're right! gonna remove that check 👍

se7entyse7en:
done

se7entyse7en:
Tested locally and it was able to detect the breaking change 👍

se7entyse7en:
Rebased and squashed.

se7entyse7en:
> Rebased and squashed.

Maybe you would like to take another look.

cmd/srcd/cmd/defered_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !integration,!regression
```
smacker: no need for this build tags anymore

se7entyse7en: Right! thank you!

se7entyse7en: done [here](https://github.com/src-d/engine/pull/363/commits/62e72fe39ea2caec81986b09976332debcdb0d41).

cmdtests/gitbase_back_comp_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+	s.PrevCmd = cmdtest.NewCommander(s.getBinary(config, tool, releases, prevVersion))
+	s.CurrCmd = cmdtest.NewCommander(s.getBinary(config, tool, releases, currVersion))
+}
+
+func (s *GitbaseBackCompTestSuite) setupWd() {
+	s.T().Helper()
+
+	_, filename, _, _ := runtime.Caller(0)
+	dir := path.Join(path.Dir(filename), "..", "..", "..")
```
smacker: are you sure it's correct path after the code was moved?

cmdtests/gitbase_back_comp_test.go
```diff
@@ -0,0 +1,274 @@
...
+	releases := regression.NewReleases("src-d", "engine", token)
+
+	s.PrevCmd = cmdtest.NewCommander(s.getBinary(config, tool, releases, prevVersion))
+	s.CurrCmd = cmdtest.NewCommander(s.getBinary(config, tool, releases, currVersion))
+}
+
+func (s *GitbaseBackCompTestSuite) setupWd() {
+	s.T().Helper()
+
+	_, filename, _, _ := runtime.Caller(0)
```
smacker: why do we need `runtime.Caller(0)` here? Isn't it equal to `os.Getwd`? If not could you please add a comment. It's not obvious at all.

cmdtests/gitbase_back_comp_test.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		s.T().Fatalf("%v", err)
+	}
+
+	homedir, err := homedir.Dir()
+	if err != nil {
+		s.T().Fatalf("unable to get home dir: %v", err)
+	}
+
+	s.testDataDir = filepath.ToSlash(filepath.Join(homedir, ".srcd", "gitbase"))
```
smacker: let's add comment that we need it only for engine <0.12

cmdtests/common.go
```diff
@@ -120,3 +105,147 @@ func (s *IntegrationTmpDirSuite) SetupTest() {
...
+	return fmt.Errorf("Different %s:\n- actual:   %v\n- expected: %v",
+		what, this, other)
+}
+
+func AreSQLOutputStrictlyEqual(s1 string, s2 string) error {
+	return ParseSQLOutput(s1).RequireStrictlyEqual(ParseSQLOutput(s2))
+}
+
+func AreSQLOutputEqual(s1 string, s2 string) error {
+	return ParseSQLOutput(s1).RequireEqual(ParseSQLOutput(s2))
```
smacker: why do you need to parse results? Is there any problem with simple comparing strings?

cmdtests/gitbase_back_comp_test.go
se7entyse7en: argh.. sorry my fault, I ran integration testing instead of regression testing before pushing 😕. I should have noticed this.

cmdtests/common.go
se7entyse7en: AFAIR it was for the rows ordering, but now I guess that we may force the query to include an `ORDER BY` clause. But I don't know whether in the future we would need to test queries strictly without `ORDER BY` clauses.

cmdtests/gitbase_back_comp_test.go
se7entyse7en: you're right, it took a couple of seconds even for me to remember the reason...

cmdtests/common.go
smacker: I don't mind to merge it as is. It just seems like an extra code that we can avoid so I mentioned it.
//cc @carlosms

cmdtests/gitbase_back_comp_test.go
se7entyse7en: addressed [here](https://github.com/src-d/engine/pull/363/commits/864cd8a628be86ce8986bb50edae6fb52f48feca).

cmdtests/gitbase_back_comp_test.go
se7entyse7en: addressed [here](https://github.com/src-d/engine/pull/363/commits/864cd8a628be86ce8986bb50edae6fb52f48feca).

cmdtests/gitbase_back_comp_test.go
se7entyse7en: addressed [here](https://github.com/src-d/engine/pull/363/commits/864cd8a628be86ce8986bb50edae6fb52f48feca).

cmdtests/gitbase_back_comp_test.go
smacker: I still don't get it. We run both integration&regression tests using `go test`
So why do you need `runtime.Caller`?
It still looks like:
```
_, filename, _, _ := runtime.Caller(0)
path.Dir(filename)
```
returns the same results as `os.Getwd`

And this new comment should belong to `setupWd` function, not `runtime.Caller`.
Or am I getting everything wrong?

se7entyse7en: > returns the same results as `os.Getwd`

It's not `path.Dir(filename)` actually, but `path.Join(path.Dir(filename), "..")`.

> So why do you need `runtime.Caller`?

Given that `regression-core` needs the test to be run with the working directory set to the root of the project, I'm exploiting the path of the current file as a reference to build the path of the root of the project. So I'm using `runtime.Caller` to get the path of `gitbase_back_comp_test.go`.

> And this new comment should belong to `setupWd` function, not `runtime.Caller`.

Yup, it actually belongs to the whole function, gonna move it 👍.

cmdtests/common.go
carlosms: :thinking: it's a shame to ask you to remove code that is already in place, but it's true that right now it's a bit convoluted.

If the only thing thing that prevents us from comparing the strings directly is the `ORDER BY`, I think it's worth it to remove this extra code, to keep the code cleaner and easier to follow.

se7entyse7en: Oh no! I remember now! But I don't know whether we want to still keep it...

There's (was) another reason. When I developed this at some point it was in-between the replacement of the SQL cli, so there was a version using our custom sql cli and another using mysql cli.

In one case the headers were all capitalized and in the other not, and also if the header is composed by multiple words the separator in one case is the space and in the other was a dash. Moreover, in our sql cli the newline was `\n` while on mysql is `\r\n`, but this latter thing is a minimal problem as we can simply replace all the newlines of the buffer.

So if we still want to test versions previous to the sql cli replacement this is needed, but I don't know whether it does make sense to test such older versions.

carlosms: I we were writing it now I would say we could ignore previous versions. Since it's already done maybe we can keep it, but let's add a comment explaining why this is done. Then later we can remove it if we don't use this to test older versions at all.

se7entyse7en:
Rebased, squashed and fixed conflicts.

se7entyse7en:
Are we ready to merge this then?

se7entyse7en:
@carlosms addressed [here](https://github.com/src-d/engine/pull/363/commits/acd153735705b97565183894614376fe9ced17a0).

PR #364, opened 2019-03-22
Attach special charactes & resize

smacker:
- The first commit sets terminal into a raw mode so special characters are propagated into container correctly.
- The second commit set correct terminal size inside a container and updates it when the size of the host terminal is changed. It solves the problem with a complex output that relies on the size of the terminal (like the one after pressing Ctrl-R in mysql-cli)
- I had to use `replace` in `go.mod` because term package of docker introduces azure/go-ansiterm dependency which depends on incorrect name of the logrus


se7entyse7en:
> The second commit set correct terminal size inside a container and updates it when the size of the host terminal is changed. It solves the problem with a complex output that relies on the size of the terminal (like the one after pressing Ctrl-R in mysql-cli)

@smacker sorry, can you please elaborate more? I didn't understand in which case this is needed.

docker/docker.go
```diff
@@ -571,5 +584,71 @@ func Attach(ctx context.Context, config *container.Config, host *container.HostC
...
+				}
+				prevH = h
+				prevW = w
+			}
+		}()
+	} else {
+		sigchan := make(chan os.Signal, 1)
+		gosignal.Notify(sigchan, signal.SIGWINCH)
+		go func() {
+			for range sigchan {
```
se7entyse7en: Shoudnt' this be as follows?
```
for {
    for range sigchan {
        resizeTty(c, containerID)
    }
}
```

docker/docker.go
```diff
@@ -571,5 +584,71 @@ func Attach(ctx context.Context, config *container.Config, host *container.HostC
...
+	ws, err := term.GetWinsize(fd)
+	if err != nil {
+		return 0, 0
+	}
+
+	return uint(ws.Height), uint(ws.Width)
+}
+
+func monitorTtySize(c *client.Client, containerID string) {
+	initTtySize(c, containerID)
```
se7entyse7en: Is this initial call needed? For windows you're actually polling, and for non-windows your waiting for a signal. Maybe eventually needed only for non-windows?

docker/docker.go
```diff
@@ -571,5 +584,71 @@ func Attach(ctx context.Context, config *container.Config, host *container.HostC
...
+				time.Sleep(10 * time.Millisecond)
+				if err = resizeTty(c, containerID); err == nil {
+					break
+				}
+			}
+		}()
+	}
+}
+
+func resizeTty(c *client.Client, containerID string) error {
```
se7entyse7en: Maybe better to change this into
```
func resizeTty(c *client.Client, containerID string, width, height uint) error
```
so that in case of windows you don't calculate `getStdOutSize` twice?

docker/docker.go
```diff
@@ -571,5 +584,71 @@ func Attach(ctx context.Context, config *container.Config, host *container.HostC
...
+				if prevW != w || prevH != h {
+					resizeTty(c, containerID)
+				}
+				prevH = h
+				prevW = w
+			}
+		}()
+	} else {
+		sigchan := make(chan os.Signal, 1)
+		gosignal.Notify(sigchan, signal.SIGWINCH)
```
se7entyse7en: Is this triggered only for this case?

docker/docker.go
smacker: It's needed for all platforms. A user can resize terminal during the start of the container.

docker/docker.go
smacker: sorry I don't understand your question. This line subscribes to SIGWINCH signal

docker/docker.go
smacker: there is no need for one more `for` and the goroutine should exit when the channel closes.
`for range` tries to read from the channel as long as it's open.

docker/docker.go
smacker: No strong opinion on it. I just think it makes code easier. The overhead is super small here. Like REALLY small.
Please let me know if you think that additional overhead is more important than code simplicity. I'll change it.

docker/docker.go
se7entyse7en: Sorry, I mean is the `SIGWINCH` signal being sent only in case of terminal resize?

docker/docker.go
se7entyse7en: Ok, I was actually concerned about the overhead 👍

docker/docker.go
se7entyse7en: What I mean is that in case of windows, if the first try in `initTtySize` fails, you then have to two goroutines doing essentially the same thing. Also if [this](https://github.com/src-d/engine/pull/364/files#diff-b715a530d535012e418a564a3eafedb9R610) is executed at a given time `t`, then it sleeps for `250ms`, in the meantime `[t, t+250]` the terminal could have been resized by the goroutine initialized [here](https://github.com/src-d/engine/pull/364/files#diff-b715a530d535012e418a564a3eafedb9R636), hence the previously calculated `prevH, prevW` could now be wrong.

docker/docker.go
smacker: you can send it manually if you want of course.
But it's how you handle terminal resize on *nix systems https://en.wikipedia.org/wiki/Signal_(IPC)#SIGWINCH

se7entyse7en: Thank you!

docker/docker.go
smacker: Those goroutines don't run at the same time. Init goroutine runs only for 50 ms and the second one waits for 250 ms. Multiple calls to `resizeTty` are safe so there is no problem with prev variables in terms of correctness of the code.

This implementation is heavily based on docker-cli implementation itself. It might have *super small* overhead in terms of trying to call resize functions when it's not really necessary. But on another hand, it's more "bullet-proof" and, in theory, should provide better UX in very weird cases.

Though I don't have a strong opinion on it. Please let me know if you still think it should be changed taking into account explanation above, I'll change.

se7entyse7en: > Those goroutines don't run at the same time. Init goroutine runs only for 50 ms and the second one waits for 250 ms.

You're right, sorry I missed that.

> Multiple calls to resizeTty are safe so there is no problem with prev variables in terms of correctness of the code.

Sorry for the noise I've just realized that you're calling the resize whenever the sizes are different from the ones detected. I got confused with the initial state. 👍

se7entyse7en: Sorry again, but I don't know why for some reason I have the impression that there's some sort of race-condition. I'm probably missing something else.

Let's say that at `t=0` the size is `(50, 50)`, then we have `(prevH, prevW) = (50, 50)`, then it sleeps for `250ms`. Let's say that during the initialization of the tty size I do a resize and, it fails 3 times, so in the end at `15ms` the current size is for example a `(100, 100)` and docker is aware of this new size hence it gives a better UX. Before waking up from the sleep of `250ms` I resize again exactly at `(50, 50)` let's say at `100ms` (the polling hasn't started yet). At `250ms` starts the polling but [this condition](https://github.com/src-d/engine/pull/364/files#diff-b715a530d535012e418a564a3eafedb9R615) is actually never met, so the UX is not optimized for the `(100, 100)` size.

smacker:
rebased on master

PR #365, opened 2019-03-22
Update latest releases URL

gomesfernanda:
Latest release URL was not leading to latest release, but for all release page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Follow these instructions:
 
 #### 2. Install source{d} Engine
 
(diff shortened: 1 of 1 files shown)


dpordomingo:
I think that in some situations `/latest` can point to a release candidate, or even worst to an old version when there is a patch for an old major.

gomesfernanda:
@marnovo what do you think about David's note?

dpordomingo:
uooops, I think you can disregard my comment.
- releases are sorted topologically, so patches to old mayors does not appear as `/latest`
- if rc are created using the `pre-release` checkbox, it is not listed as `/latest`
  - and our latest RCs in Engine did it

examples:
- https://github.com/dpordomingo/testing-repo/releases
- https://github.com/dpordomingo/testing-repo/releases/latest

Issue #330, opened 2019-03-15
Consider changing the consistency policy for gitbase mounted dirs to improve OSX performances


se7entyse7en:
For OSX changing the consistency of the mounted folder increases the performances (at least by looking at very limited testing done [here](https://github.com/src-d/engine/issues/156#issuecomment-473252486).

carlosms:
cc @src-d/data-processing in case you have any comments on this.

PR #366, opened 2019-03-22
Consistency policy mac

smacker:
Fix: #330

In this PR:
- update go.mod/sum with docker
  - commit pointing on v1.38 api version
  - manual update of some dependencies that aren't compatible anymore but go mod can't figure it out automatically
- set ConsistencyDelegated mode for volumes on macOs *always*. According to the documentation, it's the fastest mode. There the only downside => in case of container crash indexes can be written incorrectly. But gitbase handles this situation and just drops them. Also using indexes isn't the main use-case for gitbase but if they are used it's important to read them *fast*.

Before merging I also want to run some performance tests but the code is ready.


smacker:
Simple performance test (no indexes in both cases), just to validate that it works and helps:

v0.11.0:
```
$ ./srcd init ~/go/src/github.com/
$ ./srcd parse uast cmd/srcd/main.go > /dev/null // run bblfshd
$ time ./srcd sql "select * from repositories"
real	1m44.130s
user	0m0.104s
sys	0m0.081s

$ time ./srcd sql "select count(*) from repositories"
+----------+
| COUNT(*) |
+----------+
|      214 |
+----------+

real	0m0.805s
user	0m0.012s
sys	0m0.020s

$ time ./srcd sql "select count(*) from commits"
+----------+
| COUNT(*) |
+----------+
|   151778 |
+----------+

real	1m25.170s
user	0m0.012s
sys	0m0.009s
```

this branch:
```
$ ./srcd_with_policy init ~/go/src/github.com/
$ ./srcd_with_policy parse uast cmd/srcd/main.go > /dev/null // run bblfshd
$ time ./srcd_with_policy sql "select * from repositories"
real	0m50.820s
user	0m0.069s
sys	0m0.052s

$ time ./srcd_with_policy sql "select count(*) from repositories"
+----------+
| COUNT(*) |
+----------+
|      214 |
+----------+

real	0m1.975s
user	0m0.027s
sys	0m0.020s

$ time ./srcd_with_policy sql "select count(*) from commits"
+----------+
| COUNT(*) |
+----------+
|   151778 |
+----------+

real	0m35.689s
user	0m0.034s
sys	0m0.028s
```

native:
```
$ ./gitbase server -v -d ~/go/src/github.com/ -i ~/tmp/gitbase_index
<can't time it but it's super fast, like 3-5 seconds>

$ time docker run --rm --network=host mysql mysql -h 127.0.0.1 -e "select count(*) from commits"
COUNT(*)
151778

real	0m49.636s
user	0m0.039s
sys	0m0.024s
```

Results:
1. It's really much faster
2. In this branch quick queries run a bit longer due to overhead of new mysql-cli from container vs custom-go-client on host in the last release.
3. Gitbase startup time is still way slower than native (or linux) but the queries have more or less the same time.

For the documentation task I still propose to run some analyzis (like the one Fernanda is doing) on this branch and with native gitbase to indentify which queries suffer the most if any.

se7entyse7en:
Let's also decide what to do with #329.

carlosms:
I have a few questions:
- Can you explain what is going on with the docker tags? It's hard to follow. From the changes it seems `github.com/docker/docker` gets downgraded from `v1.13.1` to `v0.7.3`. And https://github.com/moby/moby (where docker/docker redirects) is already on `v17.x.x`.

- This `delegated` config is not available for linux hosts, did I get it right?

- What about `readonly` for the workdir, would that offer even faster times? Would it maybe help for linux hosts?

- About:
> There the only downside => in case of container crash indexes can be written incorrectly. But gitbase handles this situation and just drops them

What is exactly going to happen if an index is corrupted? It is ignored, and gitbase continues working as if it doesn't exist? I'm assuming the user will not be aware of this. If you try to recreate the index, will it succeed the next time?

Also regarding indexes written incorrectly. If we made the index directory mount `consistent`, would that make any difference? I mean if the user creates an index and runs `srcd stop` before the index finishes,  will it have the same index corruption as a `delegated` mount?

smacker:
> Can you explain what is going on with the docker tags

As I mentioned on slack moby repository doesn't have tags. All the tags you see there are super old and come from docker before extraction of "core" to moby. v17.x.x means 2017-xx-xx which is 2 years old.
I'm not sure why `go mod` calls it 0.7.3. It should be 0.0.0. There is no 0.7.x tags in the project. Anyway when you see `vx.x.x-0.20180524003928-df5175e1ee95` in go.mod it's a pointer to a commit not tag. More info here https://golang.org/cmd/go/#hdr-Pseudo_versions 

> This delegated config is not available for linux hosts

the is no need of using delegated mode for linux hosts. Please read docker documentation about the modes: https://docs.docker.com/docker-for-mac/osxfs-caching/
In short, it's supported right now for osx&windows, same platforms as we distribute engine on. But I chose `!= linux` instead of `== darwin || == windows` so it will start working automatically on any other platform if docker adds support to it without any changes on our side.

> What about readonly for the workdir, would that offer even faster times?

There is no mention of readonly in docker performance documentation. I don't think it's going to change anything. Though we can try.

> Would it maybe help for linux hosts?

What are you talking about? There is no problem on linux hosts.

> What is exactly going to happen if an index is corrupted?

Exactly the same thing as if container crashes right now during writing the index. Writing the index isn't atomic in gitbase. It writes to multiple files.
Current behavior I observe with broken indexes: they print warning about it on the start of gitbase, I haven't tried to recreate it. Please create a new issue if you think engine should handle this case. But it's unrelated to the PR.

>  I mean if the user creates an index and runs srcd stop before the index finishes, will it have the same index corruption as a delegated mount?

We are talking about "race condition" here. Currently, we don't test and don't handle it. Similar to the previous point. Please create a new issue if you think engine should handle this case. But it's unrelated to the PR.

carlosms:
> > Can you explain what is going on with the docker tags
> 
> As I mentioned on slack moby repository doesn't have tags. All the tags you see there are super old and come from docker before extraction of "core" to moby. v17.x.x means 2017-xx-xx which is 2 years old.
> I'm not sure why `go mod` calls it 0.7.3. It should be 0.0.0. There is no 0.7.x tags in the project. Anyway when you see `vx.x.x-0.20180524003928-df5175e1ee95` in go.mod it's a pointer to a commit not tag. More info here https://golang.org/cmd/go/#hdr-Pseudo_versions

I've been reading some more and it seems like the migration from docker/docker to moby/moby breaks a few things, and go import have been broken for years :/
The releases seem to be done in the repo github.com/docker/engine a for of moby/moby, where where docker/docker redirects...

After reading https://github.com/moby/moby/issues/33989 I've tried to use https://github.com/docker/engine imports, setting the following in `go.mod`

```
replace github.com/docker/docker => github.com/docker/engine v18.09.3
```

After a `go mod tidy` it gets replaced with
```
replace github.com/docker/docker => github.com/docker/engine v0.0.0-20190226002956-8c91e9672cc8
```

In the description you say you set manually the _commit pointing on v1.38 api version_. Where did you get that commit?

I'm thinking if the replacement of docker/docker for docker/engine works, it will hopefully allow dependabot to handle the version updates.

smacker:
> Where did you get that commit?

In git history: https://github.com/moby/moby/commit/df5175e1ee955a429ac5a4f8bdfacffd4d066c26#diff-c52a35abd674173ebde0a817d9e4e784

smacker:
btw I don't think use dependabot for docker is a good idea. Locking engine on the very last version of docker seems strange. I personally not always use the last version of docker which means I won't be able to run engine.

smacker:
we had small chat in slack to understand how to lock docker dependency and what we actually need to lock. As the result:

1. I have update commit hash to match v18.06.3-ce
2. Go mod transforms it automatically to commit hash which this PR won't fix and I'm not sure we need to fix it at all
3. We don't need dependabot for this because we want to be compatible with older docker too.
4. I added a comment in `go.mod` to clarify where the commit hash comes from
5. Engine doesn't care much about docker version because it reflects docker server version, we care about docker API version. But moby/docker doesn't provide such package separately (like we do with lookout-sdk vs lookout). The [old api client](https://github.com/docker/engine-api/) lives inside moby now.
6. Locking to some version is good enough because we use only client part of docker which is mostly auto generated code from swagger. We can update it manually when we need new feature from newer api or something like that.

carlosms:
Thanks for the summary of the slack chat, that will help us next time we need to update the api version :+1:

> > What about readonly for the workdir, would that offer even faster times?
> 
> There is no mention of readonly in docker performance documentation. I don't think it's going to change anything. Though we can try.
> 
> > Would it maybe help for linux hosts?
> 
> What are you talking about? There is no problem on linux hosts.

What I was wondering is that maybe readonly brings performance improvement in linux also. But now that I think of it, maybe we should be setting the workidr volume as readonly anyway.
Gitbase does not write into it, so it makes sense to give as little permissions as needed. Not related to this issue, we can discuss it in slack.

> > What is exactly going to happen if an index is corrupted?
> 
> Exactly the same thing as if container crashes right now during writing the index. Writing the index isn't atomic in gitbase. It writes to multiple files.
> Current behavior I observe with broken indexes: they print warning about it on the start of gitbase, I haven't tried to recreate it. Please create a new issue if you think engine should handle this case. But it's unrelated to the PR.
> 
> > I mean if the user creates an index and runs srcd stop before the index finishes, will it have the same index corruption as a delegated mount?
> 
> We are talking about "race condition" here. Currently, we don't test and don't handle it. Similar to the previous point. Please create a new issue if you think engine should handle this case. But it's unrelated to the PR.

What I wanted to clarify with those questions is that we are not introducing a new problem with a weaker consistency policy. So from your answers I see that it can break in a similar way to a `consistent` config, with maybe worse time margin between index creation and safe container stop. I think it's reasonable :+1:

Issue #372, opened 2019-03-26
Engine should remember init arguments if daemon dies


smacker:
The context is here: https://github.com/src-d/engine/issues/369

We agreed to implement it the simplest way by saving "init" argument into a file and `init` would always override this file.

PR #376, opened 2019-03-27
Engine state

smacker:
Fix: #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ This change has 2 implications:
 
 - Fix the gRCP `ResourceExhausted` error for `srcd parse uast` with big files ([#271](https://github.com/src-d/engine/issues/271)).
 - In cases where the workdir contains a lot of Git repositories, `srcd web sql` could open the web UI before gitbase was ready to accept queries ([#284](https://github.com/src-d/engine/issues/284)).
+- If the daemon container crashed, or was stopped directly by the user, the following `srcd` commands are able to restart it. But then it would not honor the working directory used when `srcd init /path/workdir` was called, using instead the current directory. Now Engine stores a file in `~/.srcd/` to solve this ([#372](https://github.com/src-d/engine/issues/372)).
 
 </details>
 
--- a/cmd/srcd/daemon/daemon.go
+++ b/cmd/srcd/daemon/daemon.go
@@ -118,12 +118,17 @@ func (o *startOptions) Save() error {
 }
 
 func Start(workdir string) error {
-	opts := startOptions{WorkDir: workdir}
+	cfg, err := config.Config()
+	if err != nil {
+		return err
+	}
+
+	opts := startOptions{WorkDir: workdir, Config: cfg}
 	if err := opts.Save(); err != nil {
 		return err
 	}
 
-	_, err := start(opts.WorkDir)
+	_, err = start(opts)
 	return err
 }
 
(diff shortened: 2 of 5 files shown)


smacker:
I'm investigating why CI fails. Tests pass locally.

smacker:
Found the issue and fixed. Though travis doesn't work today so I can't re-test it

cmd/srcd/daemon/daemon.go
```diff
@@ -74,8 +77,16 @@ func CleanUp() error {
 	}
 
 	gitbaseIndexDir := filepath.Join(datadir, "gitbase")
+	if err := os.RemoveAll(gitbaseIndexDir); err != nil {
+		return err
+	}
+
+	stateFile := filepath.Join(datadir, stateFileName)
+	if err := os.RemoveAll(stateFile); err != nil {
```
dpordomingo: if it's used, the rest is not needed.
```suggestion
	return os.RemoveAll(stateFile);
```

se7entyse7en:
LGTM, with some comments, but not blocking this PR.

In general, by reading the flow of the operations we do to ensure that the daemon is running, I have the impression that to get from A to B we first pass through C.

For example, I'd expect `ensureStarted` to first of all check whether the daemon is running and return it. But we first check for the state file, if it doesn't exist we run `Start` that saves the state (but it shouldn't be necessary if the daemon is already running as the current state file should be up-to-date), then we finally call `InfoOrStart` that returns the running container. Additionally given that `Start` just ignores the first return value `InfoOrStart`, when we call `ensureStarted` we end up calling `docker.Info` twice.

Analogously when the state file is present, we decode the state file, and then we ignore it if the daemon is already running.

smacker:
thanks Marvin. I better fix it in this PR. At least partially.

smacker:
1. I have fixed the mentioned problem in the flow. It still can run `info` more than once in some cases (like we use `docker.InfoOrStart` in `start`) but it's a very cheap operation and doesn't introduce logical complexity as previous flow.
2. Rebased on master

carlosms:
There is a small change in how `ensureStarted` works.
It is a bit of a corner case, but before you could do `srcd stop`, `cd` to another repo dir, run `srcd sql "..."` and it would query on the new dir.
Now it will open the last workdir.

I'm not worried about this specific use case, because you can run `init` explicitly, and that's how it is documented. But I'm thinking out loud: should differentiate a daemon crash to a `stop`? I mean deleting the state on `stop`. Maybe there is some other use case that relies on that.

smacker:
Could you please explain your concern in more details? I don't really understand it.

How it works on master:
init dirA ->
Everything ok case: sql query is executed on dirA, no matter did you do `cd` or not
Stop or restart docker or restart OS or some crash: sql query runs on current directory

How it works with this pr:
init dirA ->
Everything ok case: sql query is executed on dirA, no matter did you do `cd` or not
Stop or restart docker or restart OS or some crash: sql query is executed on dirA, no matter did you do `cd` or not

To change the directory you need to run `init`.

smacker:
Ah. I think I got what you mean. Difference between `stop` and `prune`.
This PR makes similar to docker. `stop` in docker doesn't remove container/state. `rm`/`prune` does.
We may use `stop` in engine as `stop&rm` in docker though, I don't have an opinion on it.

carlosms:
What I mean is that in master these two are equivalent:
```
cd /tmp/a
srcd init
srcd stop

cd /tmp/b
srcd init
srcd sql "select * from repositories"
# shows repos in /b
```

```
cd /tmp/a
srcd init
srcd stop

cd /tmp/b
srcd sql "select * from repositories"
# shows repos in /b
```

After this PR, in the second sequence of commands `srcd sql` will restart the daemon with the latest workdir, `/tmp/a`, and show the list of repos in `/a`.
As I said the specific use case is not so important to me, but still I wanted to discuss it in case there are some other use cases affected that are more important to support.

se7entyse7en:
Regarding the `stop` behavior I don't have a strong opinion, I think they're both ok until it is clearly documented.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ This change has 2 implications:
 
 - Fix the gRCP `ResourceExhausted` error for `srcd parse uast` with big files ([#271](https://github.com/src-d/engine/issues/271)).
 - In cases where the workdir contains a lot of Git repositories, `srcd web sql` could open the web UI before gitbase was ready to accept queries ([#284](https://github.com/src-d/engine/issues/284)).
+- Engine doesn't loose configuration passed in init on daemon restarts #372
```
carlosms: I suggest to add more details, only with the issue linked is not so easy to follow what it is about.

```suggestion
- If the daemon container crashed, or was stopped directly by the user, the following `srcd` commands are able to restart it. But then it would not honor the working directory used when `srcd init /path/workdir` was called, using instead the current directory. Now Engine stores a file in `~/.srcd/` to solve this ([#372](https://github.com/src-d/engine/issues/372)).
```

smacker: fixed. thanks!

PR #377, opened 2019-03-27
Add CHANGELOG.md

carlosms:
It contains the release notes for code already on master.
I added only the releases from 2019 for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@
     Click to see more.
   </summary>
 
+### Breaking Changes
+
+In this release we have changed how Engine stores the [gitbase index](https://docs.sourced.tech/gitbase/using-gitbase/indexes) data. In previous releases we stored this data in `~/.srcd/gitbase`. From now on they will be managed as docker volumes.
+
+This change has 2 implications:
+
+- If you had created any index for your data you will need to recreate them after source{d} Engine is updated.
+- The files in `~/.srcd/gitbase` are not used anymore, and you need to clean them manually:
+
+  ```bash
+  rm -rf ~/.srcd/gitbase
+  ```
+  Please not that in Linux you may need to run the command as root.
+
 ### Components
 
 - `bblfsh/bblfshd` has been updated to [v2.12.0-drivers](https://github.com/bblfsh/bblfshd/releases/tag/v2.12.0).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+
+- `srcd prune` now removes the gitbase index data ([#352](https://github.com/src-d/engine/issues/352)).
+- More friendlier and useful error messages ([#252](https://github.com/src-d/engine/issues/252), [#258](https://github.com/src-d/engine/issues/258), [#272](https://github.com/src-d/engine/issues/272), [#291](https://github.com/src-d/engine/issues/291), [#294](https://github.com/src-d/engine/issues/294), [#295](https://github.com/src-d/engine/issues/295), [#326](https://github.com/src-d/engine/issues/326)).
+- Replace the basic REPL sql shell with a standard mysql client ([#154](https://github.com/src-d/engine/issues/154)).
+- Show the container exposed ports in the `srcd components list` output ([#300](https://github.com/src-d/engine/issues/300)).
+
+### Bug Fixes
+
+- Fix the gRCP `ResourceExhausted` error for `srcd parse uast` with big files ([#271](https://github.com/src-d/engine/issues/271)).
+- In cases where the workdir contains a lot of Git repositories, `srcd web sql` could open the web UI before gitbase was ready to accept queries ([#284](https://github.com/src-d/engine/issues/284)).
```
dpordomingo: also:
```suggestion
- In cases where the workdir contains a lot of Git repositories, `srcd web sql` could open the web UI before gitbase was ready to accept queries ([#284](https://github.com/src-d/engine/issues/284)).
- Set ConsistencyDelegated mode for volumes on macOs always to improve OSX gitbase performance. Caveat: in case of container crash, indexes can be written incorrectly, but gitbase will drop them on next init ([#330](https://github.com/src-d/engine/issues/330))
```

se7entyse7en:
> lgtm, but I like the structure proposed by https://keepachangelog.com/en/1.0.0 (from the last tfm)
>
> `Added` for new features.
> `Changed` for changes in existing functionality.
> `Deprecated` for soon-to-be removed features.
> `Removed` for now removed features.
> `Fixed` for any bug fixes.
> `Security` in case of vulnerabilities.
> Considering it, all listed under "New features" should be under "Changed"
>
> I think it's a good idea to differentiate Added from Changed because it can have implications for people already using changed features.

I like the idea of having some sort of "standard labels". But at this stage, I think that we can have our own "vocabulary", and if we want to standardize it, maybe it worth having a engineering-wide convention.

carlosms:
I have added the latest changes merged into master, aeafda0.

@dpordomingo I'm not completely sold on the other format. But I'm open to discuss it.
What I'd like to ask you is consider this PR as a continuation of the format we already use for the last few releases, but in the form of a CHANGELOG file.
And then we can open the discussion about changing the format itself as a different issue, or a conversation on slack.

PR #389, opened 2019-03-28
Update gitbase-web to v0.6.4

carlosms:


carlosms:
I tagged 0.6.3 after checking that the master build was fine from a quick look to the badge in the readme.

But then the release build failed because of a small lint problem, and the badge was pointing to travis-ci.org instead of .com... So I fixed the lint error, merged, and tagged 0.6.4.

Issue #257, opened 2019-03-04
srcd sql prints weird characters on windows


smacker:
https://user-images.githubusercontent.com/406916/53730491-63a0f000-3e78-11e9-955c-b8b7ce74c480.png

carlosms:
related: #154

smacker:
this particular bug is solved but there are still weird characters in some cases. Let's fix them in this issue. Details: https://github.com/src-d/engine/issues/379#issuecomment-477958725

PR #393, opened 2019-03-29
fix input mode on windows in attach

smacker:
fix: #257


carlosms:
Let's remove the entry in unreleased/known issues from the changelog.

PR #398, opened 2019-04-02
Repl integration testing

### LINKED ISSUE #334: Test for interactive sql REPL
Current sql tests act like a pipe, e.g.
`echo "show tables;" | ./srcd sql`

We should try to test the interactive REPL as suggested in this comment:  https://github.com/src-d/engine/pull/314#issuecomment-472017411

> Most probably we still can test exit and other stuff if we run the command as
> 
> `/usr/bin/bash -e "/path/to/srcd sql"`

se7entyse7en:
Closes #334.

I've tested many other possibilities, but I failed with each of them. All the problems derived from two things:
1. make the stdin be recognized as a char mode device [here](https://github.com/src-d/engine/blob/master/cmd/srcd/cmd/sql.go#L71),
2. keep the stdin open so that I can write queries interactively.

- I've tried running the command as `/bin/bash -e <path to srcd> sql`, but it was exiting as if it was run normally,
- I've tried using `os.Stdin` as `command.Stdin`, but it has a couple of problems:
    1. the stdin was correctly recognized as char mode device (obviously),
    2. if I don't write on `command.Stdin` before starting the command, then stdin was interpreted as closed (or reads EOF) so it just exits immediately
- I've then tried to use a `Reader` (`command.Stdin` is a `Reader`) which is channel-based so that stdin is not interpreted as closed as long as I don't close the channel. This was actually both a `Reader` and a `Writer` so that it is used as a `Reader` by the program and as a `Writer` by the test to write queries:
    1. the stdin was not interpreted as closed, so I was able to write queries after the command has been started,
    2. the stdin was not recognized as a char mode device, so it was entering in [this]([here](https://github.com/src-d/engine/blob/master/cmd/srcd/cmd/sql.go#L71)) `if`,
    3. so what I was writing was actually interpreted as if the input was piped in.
- I've tried using a `bytes.Buffer`, but it has the same effect, but it was also additionally not recognized as a char mode device.
- I've then tried to improve the previous solution and tried to hack that first read by sending an `EOF` without closing the channel. I was able to implement it by configuring the `Reader` to send an `EOF` when I write a specificied marker (in the previous solution the `EOF` was being sent only when the underlying channel was closed):
    1. it was hacky, and it wasn't working, I don't even remember what exactly...

Then I discovered that I could actually run a pseudo-terminal. By using a pseudo-terminal the stdin was correctly interpreted as a char mode device, and it was not interpreted as closed immediately.


cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+
+	command.Stdout = cr
+	command.Stderr = cr
+
+	f, err := pty.Start(command)
+	if err != nil {
+		panic(err)
+	}
+
+	mysqlCliReady := make(chan struct{})
```
smacker: why do you need a channel and goroutine here if you block right after you start the goroutine?

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+
+	go func() {
+		for s := range ch {
+			lines := strings.Split(s, "\n")
```
smacker: looks like you assume you have "full" lines in `s` always. It's not true. When you read from this channel you can get this values of `s`:

1. 
```
..mysql cli header...
mys
```
2. 
```
ql>
```

and the check `strings.HasPrefix` will fail.

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+
+	return command, f, ch
+}
+
+func (s *SQLREPLTestSuite) runInteractiveQuery(stdin *os.File, query string, ch <-chan string) string {
+	io.WriteString(stdin, query)
+
+	var res strings.Builder
+	for c := range ch {
+		fmt.Printf("Message: %s\n", c)
```
se7entyse7en: Oops... Gonna remove this.

cmd/srcd/cmd/sql_test.go
se7entyse7en: You're right, in order place I didn't assume this. Gonna fix it, thanks.

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+			}
+		}
+	}()
+
+	<-mysqlCliReady
+
+	return command, f, ch
+}
+
+func (s *SQLREPLTestSuite) runInteractiveQuery(stdin *os.File, query string, ch <-chan string) string {
```
smacker: why `stdin` is  `*os.File`? It's `io.Writer`

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+	io.WriteString(stdin, query)
+
+	var res strings.Builder
+	for c := range ch {
+		fmt.Printf("Message: %s\n", c)
+		res.WriteString(c)
+		if s.containsSQLOutput(res.String()) {
```
smacker: what would happen if it can't find SQL output in res? Wait forever?

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+			break
```
smacker: looks like you expect a full output of the command in `c`. Which isn't true right now.
It can contain only
```
+--------------+
| Table        |
+--------------+
| blobs        |
```
for example

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+	return res.String()
+}
+
+func (s *SQLREPLTestSuite) exitInteractiveAndWait(timeout time.Duration, stdin *os.File, ch <-chan string) error {
+	io.WriteString(stdin, "exit;\n")
+
+	done := make(chan struct{})
+	go func() {
+		for c := range ch {
+			if strings.Contains(c, "Bye") {
```
smacker: same problem as mentioned above

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+		res.WriteString(c)
+		if s.containsSQLOutput(res.String()) {
+			break
+		}
+	}
+
+	return res.String()
+}
+
+func (s *SQLREPLTestSuite) exitInteractiveAndWait(timeout time.Duration, stdin *os.File, ch <-chan string) error {
```
smacker: `*os.File` -> `WriteCloser`

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+			if strings.Contains(c, "Bye") {
+				done <- struct{}{}
+				// don't return in order to consume all output and let the process exit
+			}
+		}
+	}()
+
+	select {
+	case <-done:
+		stdin.Close()
```
smacker: `return stdin.Close()`

cmd/srcd/cmd/sql_test.go
```diff
@@ -5,24 +5,225 @@ package cmd
...
+			if !running {
+				done <- struct{}{}
+				return
+			}
+
+			if err != nil {
+				panic(err)
+			}
+
+			time.Sleep(timeout / 5)
```
smacker: it's not very correct. Example:
timeout = 10s
1-8s - exiting
9s - exited

but this code will never know about it and fail with error. `time.After(timeout)` is going to fire much earlier than `docker.IsRunning` would return results.

se7entyse7en: The idea here is to try to check `docker.IsRunning` a couple of times within the `timeout`. I don't know if I understood what you mean.

se7entyse7en:
I've just realized that I haven't pushed on my fork.

se7entyse7en:
@smacker all issues you pointed out should have been addressed [here](https://github.com/src-d/engine/pull/398/commits/c79e10f9c45df596d1d1baba1c6b2ce057ea9068). I've extracted a `StreamLinifier` util that is responsible of yielding lines from a stream of strings.

cmd/srcd/cmd/sql_test.go
smacker: from the signature `waitMysqlCliContainerStopped(timeout time.Duration)` I understand that function will return `nil` as long as mysqlCli stopped during `timeout`. Which is not true according to the code. It will return `nil` only during `4*timeout/5 + few ms`. Tbh I feel it is a nitpick so you can just ignore it. But in case you didn't notice this problem during implementation I point to it here.

cmd/srcd/cmd/sql_test.go
```diff
@@ -1,28 +1,218 @@
...
+	select {
+	case <-done:
+		return nil
+	case <-time.After(timeout):
+		return fmt.Errorf("timeout of %v elapsed while waiting to stop container", timeout)
+	}
+}
+
+func (s *SQLREPLTestSuite) containsSQLOutput(out string) bool {
+	sep := regexp.MustCompile(`\+[-]+\+`)
```
dpordomingo: `-` inside `[]` has a special meaning not applied because there are no other chars surrounding it.
```suggestion
	sep := regexp.MustCompile(`\+-+\+`)
```
or
```suggestion
	sep := regexp.MustCompile(`\+[\-]+\+`)
```
but I don't see the benefit of `[]`

cmd/srcd/cmd/sql_test.go
```diff
@@ -106,21 +104,22 @@ func (s *SQLREPLTestSuite) TestREPL() {
 func (s *SQLREPLTestSuite) TestInteractiveREPL() {
 	require := s.Require()
 
-	command, stdin, ch := s.runInteractiveRepl()
+	command, in, out, err := s.runInteractiveRepl()
```
smacker: 👍

cmd/test-utils/common.go
```diff
@@ -177,3 +178,57 @@ func (cr *ChannelWriter) Write(b []byte) (int, error) {
...
+			case s, ok := <-in:
+				if !ok {
+					close(out)
+					return
+				}
+
+				lines := strings.Split(sl.pending+normalizeNewLine(s), "\n")
+				sl.pending = ""
+
+				lastItem := lines[len(lines)-1]
```
smacker: You can write it simpler:

```go
for i, l := range lines {
	if i == len(lines) && l != "" {
		sl.pending = l
		break
	}
	out <- l
}
```

cmd/srcd/cmd/sql_test.go
smacker: Waiting forever maybe isn't that bad we have 10m timeout...
But just in case can you try to change any query to `select * from repositories where repository_id = 'not found'` or `select nope from repositories` and check what happens. Just to be sure.

cmd/srcd/cmd/sql_test.go
```diff
@@ -1,28 +1,218 @@
-// +build integration
```
carlosms: is this intentional? if so, why?

cmd/srcd/cmd/sql_test.go
```diff
@@ -1,28 +1,218 @@
...
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/kr/pty"
```
carlosms: Will this work on windows?

cmd/srcd/cmd/sql_test.go
carlosms: This method would improve with some comments on what is supposed to look for. Now it just looks like magic.

cmd/srcd/cmd/sql_test.go
se7entyse7en: F**k, no, not intentional. Thanks.

cmd/srcd/cmd/sql_test.go
se7entyse7en: Ouch. I guess not, by looking [here](https://github.com/kr/pty/issues/50) it seems that there's a possibility to fallback to other methods, but I haven't explored it yet. Maybe we could skip this test if `runtime.GOOS == "windows"` for now and open an issue.

cmd/srcd/cmd/sql_test.go
se7entyse7en: Nice catch! I also just discovered that `-` has a special meaning only if is in-between two other chars, that why it was working.

cmd/test-utils/common.go
se7entyse7en: thank you!

cmd/srcd/cmd/sql_test.go
se7entyse7en: Sorry, I've just seen this comment. I've included a check for `Empty set`, and I tried your query and it behaves correctly. 👍 Fixed [here](https://github.com/src-d/engine/pull/398/commits/ddd1ebc8f79e38af2d81706c9af8d5a020f96025).

cmd/srcd/cmd/sql_test.go
se7entyse7en: For this case, I think that the global test timeout it's ok.

cmd/srcd/cmd/sql_test.go
se7entyse7en: Oh ok, I understand what you mean, yeah here the number `5` is just a random number to test multiple times. BTW I simplified it and now it's much more straightforward and clear. Fixed [here](https://github.com/src-d/engine/pull/398/commits/bb507a56626322b77b07452834d641894201979c).

cmd/srcd/cmd/sql_test.go
se7entyse7en: Fixed [here](https://github.com/src-d/engine/pull/398/commits/6a16536801a877db39deb7447e2dea550540f83c).

cmd/srcd/cmd/sql_test.go
se7entyse7en: Fixed [here](https://github.com/src-d/engine/pull/398/commits/5104ed7ba46092821d95a29154e478cff6eacd35).

se7entyse7en:
Gonna fix DCO before rebasing and merging.

cmd/srcd/cmd/sql_test.go
carlosms: Please take a look into it, if possible we should support windows too in the tests.

smacker: but let's do it in a separate PR. 
I mean it makes sense to disable this test on windows right now but the proper fix (if such fix is possible) better implement in a separate pr.

carlosms: ok but let's make sure the test is properly skipped in windows, and open a new issue to keep track of it.

se7entyse7en: Here it is #411.

se7entyse7en:
Rebased and squashed.

Side note: we currently cannot use `icmd` for testing the interactive SQL REPL. See [here](https://github.com/gotestyourself/gotest.tools/issues/151).

Issue #348, opened 2019-03-19
Make integration tests work on Windows


carlosms:
We should be able to run the integration test suite on Windows 10.

I looks like appveyor lists docker support on windows, so we may want to run it as CI. I'm not sure if we will want to run it as CI for every PR, but we should at least be able to run the tests manually when we tag a release candidate.

PR #401, opened 2019-04-03
Windows integration test

smacker:
Fix: #348 

Please check commit messages for the explanation for each change.

### DIFF
--- a/run-integration-tests.bat
+++ b/run-integration-tests.bat
@@ -1,0 +1,15 @@
+:: we can't use makefile for windows because it depends on CI makefile which depends on shell
+
+:: build docker image
+cd cmd/srcd-server
+docker build -t srcd/cli-daemon -f Dockerfile ../..
+cd ../..
+
+:: compile engine-cli
+if not exist build mkdir build
+if not exist build/engine_windows_amd64 mkdir build/engine_windows_amd64
+go build -o build/engine_windows_amd64/srcd.exe ./cmd/srcd
+
+:: run tests
+set SRCD_BIN=../../../build/engine_windows_amd64/srcd.exe
+go test -parallel 1 -count 1 -tags=integration github.com/src-d/engine/cmd/srcd/cmd/ -v
--- a/cmd/srcd/cmd/init.go
+++ b/cmd/srcd/cmd/init.go
@@ -51,7 +51,7 @@ var initCmd = &cobra.Command{
 
 		info, err := os.Stat(workdir)
 		if err != nil || !info.IsDir() {
-			return fmt.Errorf("path %q is not a valid working directory", workdir)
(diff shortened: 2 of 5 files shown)


run-integration-tests.bat
```diff
@@ -0,0 +1,15 @@
...
+
+:: build docker image
+cd cmd/srcd-server
+docker build -t srcd/cli-daemon -f Dockerfile ../..
+cd ../..
+
+:: compile engine-cli
+if not exist build mkdir build
+if not exist build/engine_windows_amd64 mkdir build/engine_windows_amd64
+go build -o build/engine_windows_amd64/srcd.exe ./cmd/srcd
```
carlosms: It looks like this will build the docker image as `latest`, and the `srcd version` will be `dev`.
In linux, using `make docker-build` and `make build` we will have docker image with tag `dev-sha1` and `srcd version`  will also be `dev-sha1`.
Is this correct?

The only difference I can think of is is here https://github.com/src-d/engine/blob/master/docker/dockerhub.go#L16
Which if a small thing. On linux we will ping dockerhub for newer releases, on windows it will be skipped.

I don't think we should replicate all the src-d/ci makefile logic here, so the current code is ok for me.
But do you see any other place where this difference could be a problem for future tests?

smacker: thanks for pointing this out I didn't even know about it. I often build everything without makefile and then run tests using `go test` without makefile too. I would like to be able to do that.

Why do we need to set this `dev-sha` version for integration tests at all?
And how is it related to https://github.com/src-d/engine/pull/395 ?

carlosms: > Why do we need to set this `dev-sha` version for integration tests at all?

It's not a requirement for the testing framework, really. It's just how `make build` creates the bin, with 
`-ldflags -X main.version=$(VERSION) -X main.build=$(BUILD) -X main.commit=$(COMMIT)`.

But the tests should work the same with `dev`/`latest` or `dev-sha1`/`dev-sha1`.

> And how is it related to #395 ?

The trigger for #395 was that the warning message we print when we ask dockerhub about an image that is not yet pushed. Some tests had assertions for the exact number of log lines.
We were skipping going to dockerhub for `dev-xxx` builds, so this strict tests failed for the build of a release tag. The tests are run with a real tag version, before the docker image is pushed.

With the changes in #395 we are not really fixing a bug related to dockerhub version checking, the fix was the change in the tests to not rely on the exact number of log lines. Leaving the code to hit dockerhub for any build is just a way to avoid writing future tests that pass for `dev-xx` and then fail with the release tag.

I think we are good with the current code. I would prefer to test a `.exe` as close as possible to the real one we release. But duplicating all the logic in `make build` seems a bit overkill at this stage.

smacker: thanks for the explanation!

PR #405, opened 2019-04-04
Adds integration testing for prune --with-images command

### LINKED ISSUE #374: Integration test for `prune --with-images`
#351 added tests for the `prune` command, but it does not have any test with the `--with-images` option.

Relevant discussion from the previous PR:

> This test suite does not test the `--with-images` flag.
> If the tests are run with `make test-integration` the daemon image will be the one build locally (e.g. srcd/cli-daemon:dev-b72f1fe), and deleting this image would make all the other tests fail.

> > also, there is no way to test --with-images?
> > maybe in a different Travis stage?
> 
> Maybe... or it could be the same travis stage, but a separate command we can run after `make test-integration`. Like `WITH_IMAGE_PRUNE=true make test-integration-no-build`. This would allow to skip the prune test when you run it locally. Otherwise it can be inconvenient to have the tests removing the images every time you run them on your machine.

se7entyse7en:
Closes #374.

Adds integration testing for the `srcd prune --with-images` and runs it on CI. Running this test is disabled by default and can be enabled by settings the env var `TEST_PRUNE_WITH_IMAGE=true` as in [here](https://github.com/src-d/engine/pull/405/files#diff-354f30a63fb0907d4ad57269548329e3R35).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
 
     - name: 'Integration Tests Linux'
       script:
-        - make test-integration
+        - TEST_PRUNE_WITH_IMAGE=true make test-integration
     - name: 'Generated code'
       script:
         - GO111MODULE=on go mod tidy
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,15 @@ CI_REPOSITORY ?= https://github.com/src-d/ci.git
 CI_PATH ?= $(shell pwd)/.ci
 CI_VERSION ?= v1
 
+TEST_PRUNE_WITH_IMAGE ?= false
+
 MAKEFILE := $(CI_PATH)/Makefile.main
 $(MAKEFILE):
 	git clone --quiet --branch $(CI_VERSION) --depth 1 $(CI_REPOSITORY) $(CI_PATH);
 
 -include $(MAKEFILE)
 
-GOTEST_INTEGRATION = $(GOTEST) -parallel 1 -count 1 -tags=integration -ldflags "$(LD_FLAGS)"
+GOTEST_INTEGRATION = $(GOTEST) -timeout 20m -parallel 1 -count 1 -tags=integration -ldflags "$(LD_FLAGS)"
 
 OS := $(shell uname)
 
@@ -30,6 +32,7 @@ test-integration-clean:
 endif
 
 test-integration-no-build: test-integration-clean
-	$(GOTEST_INTEGRATION) github.com/src-d/engine/cmdtests/
+	TEST_PRUNE_WITH_IMAGE=false $(GOTEST_INTEGRATION) github.com/src-d/engine/cmdtests/
+	$(GOTEST_INTEGRATION) github.com/src-d/engine/cmdtests/ -run TestPruneTestSuite/TestRunningContainersWithImages
 
 test-integration: clean build docker-build test-integration-no-build
--- a/cmdtests/prune_test.go
+++ b/cmdtests/prune_test.go
@@ -4,34 +4,35 @@ package cmdtests_test
 
 import (
 	"context"
+	"os"
 	"sort"
 	"testing"
 
 	"github.com/src-d/engine/cmdtests"
+	"github.com/src-d/engine/components"
 	"github.com/src-d/engine/docker"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
 
-// This test suite does not test the `--with-images` flag.
+// This test suite by default does not test the `--with-images` flag.
 // If the tests are run with `make test-integration` the daemon image will be
 // the one build locally (e.g. srcd/cli-daemon:dev-b72f1fe), and deleting
 // this image would make all the other tests fail.
+// Running the tests with `TEST_PRUNE_WITH_IMAGE=true make test-integration`
+// will run the test with `--with-images` flag as last in order to avoid the
+// aforementioned problem.
 
 type PruneTestSuite struct {
 	cmdtests.IntegrationTmpDirSuite
 }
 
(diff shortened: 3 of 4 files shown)


.travis.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
     - name: 'Integration Tests Linux'
       script:
-        - make test-integration
+        - TEST_PRUNE_WITH_IMAGE=true make test-integration
```
smacker: maybe better run it separately after the main test suits finished to avoid re-downloading images?

se7entyse7en: What do you mean? The test of `srcd prune --with-images` command is already done as last https://github.com/src-d/engine/pull/405/files#diff-b67911656ef5d18c4ae36cb6741b7965R36.

se7entyse7en:
Sorry as maybe this discussion has been made already and I missed it, but do we really want to also remove the mysql image? The user could already have it before using engine and IMO is unexpected that it is removed.

.travis.yml
smacker: ahhhhhh. Sorry. Thanks!

Makefile
```diff
@@ -9,13 +9,15 @@ CI_REPOSITORY ?= https://github.com/src-d/ci.git
...
+TEST_PRUNE_WITH_IMAGE ?= false
+
 MAKEFILE := $(CI_PATH)/Makefile.main
 $(MAKEFILE):
 	git clone --quiet --branch $(CI_VERSION) --depth 1 $(CI_REPOSITORY) $(CI_PATH);
 
 -include $(MAKEFILE)
 
-GOTEST_INTEGRATION = $(GOTEST) -parallel 1 -count 1 -tags=integration -ldflags "$(LD_FLAGS)"
+GOTEST_INTEGRATION = $(GOTEST) -timeout 20m -parallel 1 -count 1 -tags=integration -ldflags "$(LD_FLAGS)"
```
smacker: 😞

cmd/srcd/cmd/prune_test.go
```diff
@@ -53,8 +51,8 @@ func (s *PruneTestSuite) TearDownTest() {
 	os.RemoveAll(s.testDir)
 }
 
-func (s *PruneTestSuite) TestRunningContainers() {
-	require := s.Require()
+func (s *PruneTestSuite) testRunningContainers(require *require.Assertions, pruneCmd func() (*bytes.Buffer, error)) {
```
carlosms: Maybe it would be simpler to understand if we use `args ...string` instead of `pruneCmd func() (*bytes.Buffer, error)`.
Or it could be a `withImages bool`.
WDYT?

smacker:
> Sorry as maybe this discussion has been made already and I missed it, but do we really want to also remove the mysql image? The user could already have it before using engine and IMO is unexpected that it is removed.

A user can also have bblfsh or gitbase before using engine. So I don't see much difference here.

cmd/srcd/cmd/prune_test.go
smacker: +1 to `withImages` flag.

Makefile
carlosms: What's the reasoning for the change? The new test that removes images should not take long

se7entyse7en: > The new test that removes images should not take long

In fact it wasn't due to this new test. But locally after pruning all the images running the tests took more that 10 minutes and the process was killed 😞. The value of `20m` it's just a "random" value, maybe `15m` is enough, but I don't think it makes much difference.

cmd/srcd/cmd/prune_test.go
se7entyse7en: yeah much better, [here](https://github.com/src-d/engine/pull/405/commits/f18d31fb8a9ae8bd93c8a7e30d64a21b7666f956).

cmd/srcd/cmd/prune_test.go
```diff
@@ -100,6 +104,69 @@ func (s *PruneTestSuite) TestRunningContainers() {
...
+}
+
+func (s *PruneWithoutImagesTestSuite) TestStoppedContainers() {
+	require := s.Require()
+
+	out, err := s.RunCommand(context.TODO(), "prune")
+	require.NoError(err, out.String())
+}
+
+type PruneWithImagesTestSuite struct {
```
smacker: question: why do you need another suite for `TestRunningContainersWithImages`?

se7entyse7en: Just to run it separately as last suite in the Makefile. Eventually we could have a single suite and then change the Makefile to run the specific test in the suite.

cmd/srcd/cmd/prune_test.go
```diff
@@ -100,6 +104,69 @@ func (s *PruneTestSuite) TestRunningContainers() {
...
+	PruneTestSuite
+}
+
+func (s *PruneWithImagesTestSuite) TestRunningContainersWithImages() {
+	if os.Getenv("TEST_PRUNE_WITH_IMAGE") != "true" {
+		s.T().Skip("Skipping testing pruning with --with-images flag")
```
carlosms: Suggestion to make it clear how to run without having to dig into the code:
```suggestion
		s.T().Skip("Use env var TEST_PRUNE_WITH_IMAGE=true to test srcd prune with --with-images flag")
```

cmd/srcd/cmd/prune_test.go
smacker: sorry, could you explain more your idea? I didn't really get it. Why can't you run `PruneTestSuite/TestRunningContainersWithImages` in makefile?

se7entyse7en: That's actually the same thing I said 😄. We can have a single suite and two different tests and run `PruneTestSuite/TestRunningContainersWithImages` in the Makefile 👍. Gonna do it.

se7entyse7en:
Rebased and squashed, gonna merge when all green.

Issue #378, opened 2019-03-27
Wait for index creation properly instead of waiting an arbitrary amount of time


se7entyse7en:
In some tests we create indexes, but instead of waiting for their creation we simply sleep a small amount of time. We should have a more robust mechanism than this.

Some context [here](https://github.com/src-d/engine/pull/363#discussion_r269600489).

PR #407, opened 2019-04-04
integration tests: wait for index creation 

smacker:
Fix: #378

### DIFF
--- a/cmdtests/sql_index_util.go
+++ b/cmdtests/sql_index_util.go
@@ -1,0 +1,55 @@
+// +build integration regression
+
+package cmdtests
+
+import (
+	"strings"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"gotest.tools/icmd"
+)
+
+// building/loading index takes some time, wait maximum 15s
+func IndexIsVisible(s commandSuite, table, name string) string {
+	var visibleValue string
+	for i := 0; i < 15; i++ {
+		visibleValue = hasIndex(s, table, name)
+		if visibleValue != "YES" {
+			time.Sleep(time.Second)
+		} else {
+			break
+		}
+	}
+
+	return visibleValue
+}
+
+func hasIndex(s commandSuite, table, name string) string {
+	r := s.RunCommand("sql", "SHOW INDEX FROM "+table)
+	s.Require().NoError(r.Error, r.Combined())
+
+	// parse result and check that correct index was built and it is visiable
+	lines := strings.Split(r.Stdout(), "\n")
+	for _, line := range lines {
+		cols := strings.Split(line, "|")
+		if len(cols) < 15 {
+			continue
+		}
+		if strings.TrimSpace(cols[1]) != table {
+			continue
+		}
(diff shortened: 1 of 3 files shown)


cmdtests/sql_test.go
```diff
@@ -224,26 +222,27 @@ func (s *SQLTestSuite) TestIndexesWorkdirChange() {
...
-
 	if hasIndex {
-		// parse result and check that correct index was built and it is visiable
-		indexLine := strings.Split(r.Stdout(), "\n")[3]
-		expected := `repositories.repository_id`
-		require.Contains(indexLine, expected)
-		visibleValue := strings.TrimSpace(strings.Split(indexLine, "|")[14])
+		var visibleValue string
+		// building/loading index takes some time, wait maximum 15s
+		for i := 0; i < 15; i++ {
```
dpordomingo: For the simple index we are creating in this testcase, 15s can be enough.
with bigger datasets (as could be needed in the future), it would not be enough.

smacker: 15s is a lot. I would prefer it to fail when we add such a big dataset. And increase timeout/reduce dataset depends on the case.

se7entyse7en:
> I have found only one place where we were checking index only with 1s sleep.

Yup the only other place is still unmerged in #363.

smacker:
Ok. I'll wait for it to be merged and then rebase.

smacker:
I have reworked the PR due to different check in regression test. Please another pass.

cmdtests/sql_index_util.go
```diff
@@ -0,0 +1,55 @@
...
+			time.Sleep(time.Second)
+		} else {
+			break
+		}
+	}
+
+	return visibleValue
+}
+
+func hasIndex(s commandSuite, table, name string) string {
```
se7entyse7en: Does the visible column have other values than "YES" and "NO"?

smacker: according to mysql documentation no. At least there is no mention of any other value.

se7entyse7en: I'd then change the return value to `bool` so that the handling of "YES" and "NO" values is done only within this function.

se7entyse7en:
LGTM 👍

cmdtests/sql_index_util.go
```diff
@@ -0,0 +1,55 @@
...
+
+	return hasIndex(s, table, name)
+}
+
+func hasIndex(s commandSuite, table, name string) bool {
+	r := s.RunCommand("sql", "SHOW INDEX FROM "+table)
+	s.Require().NoError(r.Error, r.Combined())
+
+	// parse result and check that correct index was built and it is visiable
+	lines := strings.Split(r.Stdout(), "\n")
```
carlosms: Maybe we could add an example output. This way if the format changes in the future we can compare it and fix the code easier.

smacker: I added a link instead of output itself (it's huge). The link contains version of mysql so it won't change.

PR #409, opened 2019-04-04
Adds an utility to print tables

### LINKED ISSUE #322: Inconsistent style in tables
These two commands should have the same style for the table header

```
$ go run cmd/srcd/main.go components list
IMAGE                             INSTALLED    RUNNING    CONTAINER NAME
bblfsh/bblfshd:v2.11.8-drivers    yes          yes        srcd-cli-bblfshd
bblfsh/web:v0.9.0                 no           no         srcd-cli-bblfsh-web
srcd/cli-daemon:latest            yes          yes        srcd-cli-daemon
srcd/gitbase-web:v0.6.2           no           no         srcd-cli-gitbase-web
srcd/gitbase:v0.19.0              yes          no         srcd-cli-gitbase
```

```
$ go run cmd/srcd/main.go parse drivers list
LANGUAGE	VERSION
----------	----------
python		v2.8.0
cpp		v1.1.0
java		v2.5.0
javascript	v2.6.0
bash		v2.4.0
ruby		v2.9.0
go		v2.5.0
csharp		v1.4.0
php		v2.7.0
```

se7entyse7en:
Closes #322.

This adds two functions `PrintTable` and `PrintDefaultTable`. This simplifies table printing and helps to have a consistent styling for all of them. It's also easier to eventually change the style of all tables.

### DIFF
--- a/cmd/srcd/cmd/components.go
+++ b/cmd/srcd/cmd/components.go
@@ -20,7 +20,6 @@ import (
 	"log"
 	"os"
 	"strings"
-	"text/tabwriter"
 
 	"github.com/spf13/cobra"
 	"github.com/src-d/engine/components"
@@ -47,11 +46,10 @@ var componentsListCmd = &cobra.Command{
 			return humanizef(err, "could not list images")
 		}
 
-		w := tabwriter.NewWriter(os.Stdout, 0, 0, 4, ' ', 0)
-		fmt.Fprintf(w, "IMAGE\tINSTALLED\tRUNNING\tPORT\tCONTAINER NAME\n")
-
+		t := NewTable("%s", "%s", "%v", "%v", "%v")
+		t.Header("IMAGE", "INSTALLED", "RUNNING", "PORT", "CONTAINER NAME")
 		for _, cmp := range cmps {
-			fmt.Fprintf(w, "%s\t%s\t%v\t%v\t%v\n",
+			t.Row(
 				cmp.ImageWithVersion(),
 				boolFmt(cmp.IsInstalled()),
 				boolFmt(cmp.IsRunning()),
@@ -60,8 +58,7 @@ var componentsListCmd = &cobra.Command{
 			)
 		}
 
-		w.Flush()
-		return nil
+		return t.Print(os.Stdout)
 	},
 }
 
--- a/cmd/srcd/cmd/table.go
+++ b/cmd/srcd/cmd/table.go
@@ -7,20 +7,27 @@ import (
 	"text/tabwriter"
 )
 
+// Table represents a printable table composed by rows and an optional header
 type Table struct {
 	formats []string
 	header  []string
 	rows    [][]interface{}
 }
 
+// Header sets the header of the table with the given strings
 func (t *Table) Header(header ...string) {
 	t.header = header
 }
 
+// Row adds a row to the table
 func (t *Table) Row(row ...interface{}) {
 	t.rows = append(t.rows, row)
 }
 
+// Print prints the table to the given writer
+// It returns an error if there's a mismatch between the length of the formats
+// and the length of the header, or between the length of the formats and the
+// length of any row.
 func (t *Table) Print(output io.Writer) error {
 	tw := tabwriter.NewWriter(output, 0, 0, 4, ' ', 0)
 	if len(t.header) > 0 {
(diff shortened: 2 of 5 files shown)


cmd/srcd/cmd/table.go
```diff
@@ -0,0 +1,29 @@
...
+
+	sFormats := strings.Join(formats, "\t") + "\n"
+	for _, row := range data {
+		fmt.Fprintf(tw, sFormats, row...)
+	}
+
+	return tw.Flush()
+}
+
+func PrintDefaultTable(output io.Writer, header []string, formats []string, data [][]interface{}) error {
```
smacker: what is the use case for `PrintTable`?

se7entyse7en: Currently none 😅. So yes, I guess that we can simply rename `PrintDefaultTable` into `PrintTable`  and remove the current one.

cmd/srcd/cmd/components.go
```diff
@@ -47,21 +46,21 @@ var componentsListCmd = &cobra.Command{
 			return humanizef(err, "could not list images")
 		}
 
-		w := tabwriter.NewWriter(os.Stdout, 0, 0, 4, ' ', 0)
-		fmt.Fprintf(w, "IMAGE\tINSTALLED\tRUNNING\tPORT\tCONTAINER NAME\n")
+		header := []string{"IMAGE", "INSTALLED", "RUNNING", "PORT", "CONTAINER NAME"}
+		formats := []string{"%s", "%s", "%v", "%v", "%v"}
+		data := make([][]interface{}, len(cmps))
```
smacker: just an idea of a little bit different API to avoid types convertions:
```go
table := table.New("%s", "%s", "%v", "%v", "%v")
table.Headers("IMAGE", "INSTALLED", "RUNNING", "PORT", "CONTAINER NAME")
for _, cmp := range cmps {
	table.Row(
		cmp.ImageWithVersion(),
		boolFmt(cmp.IsInstalled()),
		boolFmt(cmp.IsRunning()),
		publicPortsFmt(cmp.GetPorts()),
		cmp.Name,
	)
}

return table.Print(os.Stdout)
```

se7entyse7en: I like it! It's much cleaner than dealing with that `[][]interface{}`.

se7entyse7en:
@smacker done, I force-pushed as changing that api actually changed all the commits and the PR is small.

cmd/srcd/cmd/table.go
```diff
@@ -0,0 +1,53 @@
...
+	"text/tabwriter"
+)
+
+type Table struct {
+	formats []string
+	header  []string
+	rows    [][]interface{}
+}
+
+func (t *Table) Header(header ...string) error {
```
smacker: you don't check for this error in the code above that uses this method. Imo it would make API not so pretty if you do check. 
I would suggest to use `sticky error` pattern here. One of the examples of it is [bufio.Scanner](https://golang.org/pkg/bufio/#Scanner) from stdlib. But instead of defining `Err() error` method we can return it from `Print()`.

I have also found a very recent post from Dave Cheney explaining exactly this! https://dave.cheney.net/2019/01/27/eliminate-error-handling-by-eliminating-errors

cmd/srcd/cmd/table.go
```diff
@@ -0,0 +1,53 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"text/tabwriter"
+)
+
+type Table struct {
```
carlosms: We have too many golint warnings about missing docs for exported types and functions.
Let's start to document them please.

cmd/srcd/cmd/table.go
se7entyse7en: thanks for the tip!

se7entyse7en: Addressed [here](https://github.com/src-d/engine/pull/409/commits/2a2e1f06b72c18c2130c3a678ca254f9dd7b1984).

smacker: It doesn't implement the pattern but in this case most probably it doesn't matter anyway and actually simplifies the code 👍

cmd/srcd/cmd/table.go
se7entyse7en: Added [here](https://github.com/src-d/engine/pull/409/commits/7e03da468499a68ae348b8636b3dec0c44c2e09a).

cmd/srcd/cmd/table.go
se7entyse7en: Yup, in this case, I think that this is enough :), but I'll surely remember this pattern for more complex things.

PR #425, opened 2019-04-09
Migrate to src-d/go-cli and src-d/go-log

### LINKED ISSUE #410: Migrate from cobra to src-d/go-cli
Other src-d projects use [github.com/src-d/go-cli](https://github.com/src-d/go-cli/).
It makes sense to have a unified way of doing things from the developers perspective, and also to provide a more unified user experience.

Related discussion: https://github.com/src-d/engine/issues/244

carlosms:
Fix #410.

This PR modifies both `srcd` and `srcd-server` commands to use https://github.com/src-d/go-cli, and replaces logrus with https://github.com/src-d/go-log.

src-d/go-cli always adds a `version` sub command by default. Since engine already had a different `version`, I created https://github.com/src-d/go-cli/pull/15. The `go.mod` in this PR uses my fork.
- [x] TODO: change `go.mod` after https://github.com/src-d/go-cli/pull/15 is merged.

The most tricky part was `defered` and `defered_test`. I don't really like how it turned out, but I can't think of anything better right now.
Also, you'll see I added a couple of new checks in `Defered.spinner`.
- Detect log level manually. Since we are writing to stderr directly, this cannot rely on the log library skipping this print if the level is `error` for example.
- Detect if the output is a terminal or not. If you do `srcd sql "select 1" 2> /tmp/out.txt` in master you'll see that we are doing the spinner in the text file. I think this shouldn't happen, so I fixed it.

Before merging I also want to add this to the changelog:
- [x] replace `-v/--verbose` with `--log-level`, list new options for log level
- [x] srcd/cli-daemon image now requires `serve` sub command.


cmd/srcd/cmd/components.go
```diff
@@ -17,49 +17,49 @@ package cmd
...
 }
 
 // componentsListCmd represents the components list command
-var componentsListCmd = &cobra.Command{
-	Use:   "list",
-	Short: "List source{d} components",
-	RunE: func(cmd *cobra.Command, args []string) error {
-		allVersions, _ := cmd.Flags().GetBool("all")
+type componentsListCmd struct {
+	Command `name:"list" short-description:"List source{d} components" long-description:"List source{d} components"`
```
smacker: maybe avoid repeating the same description? I'm not sure where `go-cli` uses `long-description`. If it's necessary for the nice output we can PR usage of short one if long is empty.

carlosms: Both are needed:
```$ srcd -h
Usage:
  srcd [OPTIONS] <command>

Available commands:
  version     Short goes here

$ srcd version -h
Usage:
  srcd [OPTIONS] version [version-OPTIONS]

Long goes here
```

I agree with your idea to use `short` when `long`  is not set. But I'd prefer if you could open an issue in https://github.com/src-d/go-cli instead of blocking this PR.

cmd/srcd/cmd/defered.go
```diff
@@ -24,11 +25,27 @@ type Defered struct {
...
 	SpinnerInterval time.Duration
+
+	// logger is the go-log DefaultLogger. Can be changed for tests
+	logger log.Logger
+	// logWriter is the io.Writer where logger sends the output. By default,
+	// os.Stderr. Can be changed for tests
+	logWriter io.Writer
+	// forceIsTerminal forces spinner to be printed without checking if the
+	// output is a terminal. For tests
+	forceIsTerminal bool
```
smacker: maybe we can rename it to `isTerminal` and set in the constructor.
For the tests you can do `d := newDefered(); d.isTerminal = true/false` or create this struct without a constructor.
Imo it will be cleaner code.

cmd/srcd/cmd/defered.go
```diff
@@ -24,11 +25,27 @@ type Defered struct {
...
 }
 
 // Print function would print message after timeout
 // returns cancel function that should be called to cancel printing
 func (d *Defered) Print() func() {
+	if d.logger == nil {
```
smacker: also can be set in a constructor to avoid `if`s

cmd/srcd/cmd/defered_test.go
```diff
@@ -1,47 +1,52 @@
...
+	cancel := d.Print()
+	time.Sleep(timeout)
+	cancel()
+	s.mockLogger.Infof("End")
 }
 
 func (s *DeferedTestSuite) buildDefered(withSpinner bool, inputFn func(stop <-chan bool) <-chan string) *Defered {
 	s.T().Helper()
 
+	l := mockLogger{}
```
smacker: nit pick. You don't need this variable. You can do `s.mockLogger = &mockLogger{}` and `logger: s.mockLogger` below

cmd/srcd/cmd/root.go
```diff
@@ -17,53 +17,30 @@ package cmd
...
+// other application commands to provide default behavior for logging, config
+type Command struct {
+	cli.PlainCommand
+	cli.LogOptions `group:"Log Options"`
 
-// rootCmd represents the base command when called without any subcommands
-var rootCmd = &cobra.Command{
-	Use:           "srcd",
-	Short:         "The Code as Data solution by source{d}",
-	SilenceErrors: true,
```
se7entyse7en: How does it behave now in case of error?

cmd/srcd/config/config.go
```diff
@@ -1,83 +1,71 @@
...
 
-	checkConfig()
+	return checkConfig()
 }
 
-func checkConfig() {
-	actualConfigFile := viper.ConfigFileUsed()
-	if actualConfigFile == "" {
-		return
+func checkConfig() error {
```
se7entyse7en: This is called only [here](https://github.com/src-d/engine/pull/425/files#diff-a727f8a7dbe085385785155fa0679272R32). I'd remove this, and move the logic inside `InitConfig`.

cmd/srcd/config/config.go
```diff
@@ -1,83 +1,71 @@
...
-			actualConfigFile))
+		return errors.Wrapf(err, "Error checking config file '%s'", configFile)
 	}
+
+	return nil
 }
 
-// Config returns the config used (from a file, env, or defaults)
+// Config returns the config used
 func Config() (*api.Config, error) {
```
se7entyse7en: This is called only [here](https://github.com/src-d/engine/blob/87b3d351b370d6d41065a64536d2afb9d4c6a6c7/cmd/srcd/daemon/daemon.go#L141), but it maybe called multiple times given that we may call `ensureStarted` more than once. So we actually unmarshall the configuration every time we need it.

What about having a global `var conf api.Config`? I'd unmarshall the config only once at `InitConfig` and then add functioin to get that config.

cmd/srcd/cmd/defered.go
```diff
@@ -54,43 +72,60 @@ func (d *Defered) print(ch <-chan bool, done chan<- bool) {
...
+
 	interval := d.SpinnerInterval
 	if interval == 0 {
 		interval = 200 * time.Millisecond
 	}
 
-	writer := logrus.StandardLogger().Out
+	// If the logger format is not text, or the output is not a terminal,
+	// do not print the spinner
+	if log.DefaultFactory.Format != log.TextFormat || !d.isTerminal() {
```
se7entyse7en: 👍

cmd/srcd/cmd/defered.go
```diff
@@ -54,43 +72,60 @@ func (d *Defered) print(ch <-chan bool, done chan<- bool) {
...
 			return
 		default:
 			spinner := string(charset[i%len(charset)])
-			logrus.Infof("%s %s", d.Msg, spinner)
+			d.logger.Infof("%s %s", d.Msg, spinner)
 			time.Sleep(interval)
-			if writer == os.Stdout || writer == os.Stderr {
-				fmt.Fprint(writer, "\033[A")
-			}
+			fmt.Fprint(d.logWriter, "\033[A")
```
se7entyse7en: Would it be possible to make `go-log` support these kinds of sequences? We could remove both [this](https://github.com/src-d/engine/pull/425/files#diff-0683c59e10bb2d32b77d1f043eb9b4dbR91) and `d.logWriter`.

api/config.go
```diff
@@ -1,8 +1,10 @@
 package api
 
 import (
-	"github.com/sirupsen/logrus"
+	fmt "fmt"
```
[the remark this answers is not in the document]
carlosms: didn't notice, it was added by the editor. Changed in f6e580c.

cmd/srcd/cmd/defered.go
carlosms: Done in a338701

cmd/srcd/cmd/root.go
carlosms: I missed this, I thought we didn't change this in master. It is actually not silenced in master:
```
$ srcd init /tmp/nope
Usage:
  srcd init [workdir] [flags]

Flags:
  -h, --help   help for init

Global Flags:
      --config string   config file (default is $HOME/.srcd/config.yml)

path '/tmp/nope' is not a valid working directory
```

In this PR works as expected:
```

$ srcd init /tmp/nope
path '/tmp/nope' is not a valid working directory

$ srcd init --nope
unknown flag `nope'
Usage:
  srcd [OPTIONS] init [init-OPTIONS] [workdir]

Starts the daemon or restarts it if already running

Help Options:
  -h, --help                                     Show this help message

[init command options]
          --config=                              config file (default: $HOME/.srcd/config.yml)

    Log Options:
          --log-level=[info|debug|warning|error] Logging level (default: info) [$LOG_LEVEL]
          --log-format=[text|json]               log format, defaults to text on a terminal and json otherwise [$LOG_FORMAT]
          --log-fields=                          default fields for the logger, specified in json [$LOG_FIELDS]
          --log-force-format                     ignore if it is running on a terminal or not [$LOG_FORCE_FORMAT]
```

cmd/srcd/config/config.go
carlosms: done in f164c20

cmd/srcd/config/config.go
carlosms: I have refactored the file in 51122b2, I think now it's closer to what you describe.

cmd/srcd/cmd/defered.go
carlosms: I don't think this should be part of go-log. In a way I feel we are misusing logs to show program output.

se7entyse7en: fair enough, yup this isn't really a "real" log.

carlosms:
Rebased on top of master to fix conflicts and clean the history.

@smacker @dpordomingo it's getting a bit hard to keep up with changes in master, I'd appreciate if you could prioritize this one when you allocate some time to review code.

PR #428, opened 2019-04-10
Adds checks when building Docker client

### LINKED ISSUE #417: Handle missing docker desktop or docker toolbox installation.
In #415 instead of failing with a panic, srcd should have been able to detect the required docker desktop installation, and fail gracefully.

Scenarios I can think of:
- Docker version is lower than we require
- Docker toolbox is found
- Docker is missing completely

se7entyse7en:
Closes #417.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -28,8 +28,44 @@ import (
 
 type Port = types.Port
 
+// GetClient returns a docker client if all checks pass.
+// This function performs three checks:
+//   1. checks that docker is installed and running properly,
+//   2. checks that the user is not running docker toolbox.
+//   3. checks that the client api version is supported by the docker engine,
+func GetClient() (*client.Client, error) {
+	logrus.Debug("Creating docker client from env")
+	// This will fail in case of bad response from the daemon or in
+	// case of docker not installed/running
+	c, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Debug("Checking for Docker Toolbox")
+	var info types.Info
+	// Get information from running daemon to check whether is running
+	// docker toolbox
+	info, err = c.Info(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.Contains(strings.ToLower(info.OperatingSystem), "boot2docker") {
+		return nil, fmt.Errorf("Docker Toolbox is not supported")
+	}
+
+	logrus.Debug("Retrieving docker server version")
+	// Call `ServerVersion` to force checking API version compatibility
+	if _, err = c.ServerVersion(context.Background()); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
 func Version() (string, error) {
-	c, err := client.NewEnvClient()
+	c, err := GetClient()
 	if err != nil {
 		return "", errors.Wrap(err, "could not create docker client")
 	}
@@ -50,7 +86,7 @@ var ErrNotFound = errors.New("container not found")
 type Container = types.Container
 
 func Info(name string) (*Container, error) {
-	c, err := client.NewEnvClient()
+	c, err := GetClient()
 	if err != nil {
 		return nil, errors.Wrap(err, "could not create docker client")
 	}
@@ -80,7 +116,7 @@ func Info(name string) (*Container, error) {
 }
 
 func List() ([]Container, error) {
-	c, err := client.NewEnvClient()
+	c, err := GetClient()
 	if err != nil {
 		return nil, errors.Wrap(err, "could not create docker client")
 	}
@@ -131,7 +167,7 @@ func RemoveContainer(name string) error {
 		return err
 	}
 
-	c, err := client.NewEnvClient()
+	c, err := GetClient()
 	if err != nil {
 		return errors.Wrap(err, "could not create docker client")
 	}
@@ -205,7 +241,7 @@ func SplitImageID(id string) (image, version string) {
 
 // Pull an image from docker hub with a specific version.
 func Pull(ctx context.Context, image, version string) error {
-	c, err := client.NewEnvClient()
+	c, err := GetClient()
 	if err != nil {
 		return errors.Wrap(err, "could not create docker client")
 	}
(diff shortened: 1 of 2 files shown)


docker/docker.go
```diff
@@ -28,8 +28,44 @@ import (
...
+// GetClient returns a docker client if all checks pass.
+// This function performs three checks:
+//   1. checks that the user is not running docker toolbox.
+//   2. checks that docker is installed and running properly,
+//   3. checks that the client api version is supported by the docker engine,
+func GetClient() (*client.Client, error) {
+	if dockerClient == nil {
+		logrus.Debug("Checking for Docker Toolbox")
+		// If docker toolbox is installed then the environment variable
+		// `DOCKER_TOOLBOX_INSTALL_PATH` is not empty
```
smacker: I'm not sure it's a correct way to check for the docker toolbox.
Afaik you can have both toolbox&native installed. But should run only one.

se7entyse7en: > Afaik you can have both toolbox&native installed.

Ouch, didn't know that. Gonna try.

se7entyse7en: Ok I've also installed Docker Toolbox locally and I've fixed the check 👍 [here](https://github.com/src-d/engine/pull/428/commits/370dd8a197f6da8034a02988c19901233add757a).

docker/docker.go
```diff
@@ -52,6 +45,17 @@ func GetClient() (*client.Client, error) {
...
 		}
 
+		logrus.Debug("Checking for Docker Toolbox")
+		// The following `SplitN` is safe as the `c.DaemonHost()` has been
+		// already checked when creating the client
+		scheme := strings.SplitN(c.DaemonHost(), "://", 2)[0]
+		// If scheme is `tcp` it means that the host is using a VM
+		// In case of Docker Desktop for Unix this is `unix`
+		// In case of Docker Desktop for Windows this is `npipe`
+		if scheme == "tcp" {
```
smacker: this check is also incorrect. Native docker [can use](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option) tcp too, it just isn't default option.

se7entyse7en: Thank you! I've just read about all the options. Ok, by digging more I think that the only reliable information is [this](https://docs.docker.com/engine/api/v1.39/#operation/SystemInfo), in particular, the `OperatingSystem `.

In my case I have:
```
Operating System: Boot2Docker 18.09.4 (TCL 8.2.1)
```
and:
```
Operating System: Docker for Mac
```

So I guess that I could do something like `strings.Contains(os, "boot2docker")`. WDYT?

smacker: if this is the only option we can differentiate "native" docker from docker-toolbox/docker-machine.

se7entyse7en: I've tested also on Windows and the `Operating System` contains `Boot2Docker` when using Docker Toolbox and `Docker for Windows` when natively, similarly to OSX.

se7entyse7en: Addressed [here](https://github.com/src-d/engine/pull/428/commits/6399fa108da6f1216b2f38f3fe24a1262d81f2a0).

docker/docker.go
```diff
@@ -28,8 +28,47 @@ import (
 
 type Port = types.Port
 
+var dockerClient *client.Client
```
smacker: is it possible that we will call `GetClient` concurrently?

se7entyse7en: mhmmm... that's a very good point. Honestly, I don't know. But just in case I think that it is expected `GetClient` to be thread-safe. I'd add a mutex when creating the client. WDYT?

smacker: Why do you need singleton here at all? Because of the overhead? How big is it?

se7entyse7en: > Why do you need singleton here at all? Because of the overhead?

Mainly because I found a bit awkward to build the client every time.

> How big is it?

I haven't actually measured it.

se7entyse7en: I've tried [here](https://github.com/src-d/engine/pull/428/commits/41ca684153749fc32cf8e74412d8df4ce0a9e582) to make it thread-safe. Let me know wdyt and if you prefer each function to build its own client.

carlosms: I'm also not sure if a singleton is needed.
I could call `client.Close()` somewhere and break the client of other goroutine, for example.

se7entyse7en: True. I guess that it's not that safe in this case to make a singleton without actually wrapping the original client to avoid for example exposing the `Close` method or others that might break. But at least for now it seems an unneeded overkill given that the overhead shouldn't be an issue.

dpordomingo: I see neither the necessity of making the client a singleton.
In which circumstances do we need to ensure there is only one instance of the docker client?

se7entyse7en: > In which circumstances do we need to ensure there is only one instance of the docker client?

No actually need to ensure uniqueness, I just found a bit awkward to create a client every time we need it. But since it introduces more issues than it "solves" I'm gonna remove it.

se7entyse7en: Removed singleton [here](https://github.com/src-d/engine/pull/428/commits/71c5d42bc85eccb016401709ca5cac7fae7e63e4).

docker/docker.go
```diff
@@ -28,8 +28,44 @@ import (
 
 type Port = types.Port
 
+// GetClient returns a docker client if all checks pass.
+// This function performs three checks:
+//   1. checks that docker is installed and running properly,
+//   2. checks that the user is not running docker toolbox.
+//   3. checks that the client api version is supported by the docker engine,
+func GetClient() (*client.Client, error) {
```
carlosms: I'm not sure if we should we export this.
The rest of the code is calling this package convenience wrappers, not getting a client and use it directly.

For `prune_test.go` I think it makes sense to use `docker.GetClient()` directly. Or move `ListNetworks` to this package if needed.

What are others' thoughts?

smacker: it makes sense to move `ListNetworks` into docker package.
but I don't see a big problem with exporting `GetClient` also.

se7entyse7en: Agree for moving `ListNetworks`, but I'd keep `GetClient` exposed even if so far there's no need to use it from other packages.

se7entyse7en: Moved `ListNetworks` [here](https://github.com/src-d/engine/pull/428/commits/35b17e92040a95d26498a37e31d2b2d449965624).

carlosms:
In case you didn't notice, build is failing on travis.

se7entyse7en:
@carlosms thank you! I hadn't notice.

se7entyse7en:
Tests are failing due to bug solved by #432.

se7entyse7en:
Rebased and squashed, gonna merge once all green.

PR #431, opened 2019-04-10
Makes Gitbase workdir bind-mount read-only

### LINKED ISSUE #421: Mount gitbase workdir as readonly
Gitbase does not need to write into the workdir.
When we create the volume we can make it readonly, it makes sense to keep the permissions to the minimum needed.

se7entyse7en:
Closes #421.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -292,19 +292,24 @@ func WithEnv(key, value string) ConfigOption {
 }
 
 func WithVolume(name, containerPath, hostOS string) ConfigOption {
-	return withVolume(mount.TypeVolume, name, containerPath, hostOS)
+	return withVolume(mount.TypeVolume, name, containerPath, false, hostOS)
 }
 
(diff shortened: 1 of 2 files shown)


docker/docker.go
```diff
@@ -292,19 +292,28 @@ func WithEnv(key, value string) ConfigOption {
 }
 
 func WithVolume(name, containerPath, hostOS string) ConfigOption {
-	return withVolume(mount.TypeVolume, name, containerPath, hostOS)
+	return withVolume(mount.TypeVolume, name, containerPath, false, hostOS)
+}
+
+func WithROVolume(name, containerPath, hostOS string) ConfigOption {
```
smacker: I don't see why would we ever need `ROVolume`. And even if we will need it we can add it later.

se7entyse7en: Addressed [here](https://github.com/src-d/engine/pull/431/commits/4afbc44701165e9b523197e4cc8691c40428902d).

PR #433, opened 2019-04-11
Components start cmd

smacker:


cmd/srcd/cmd/components.go
```diff
@@ -145,10 +132,71 @@ var componentsInstallCmd = &cobra.Command{
 	},
 }
 
+// componentsStartCmd represents the components start command
+var componentsStartCmd = &cobra.Command{
+	Use:   "start [component]",
```
se7entyse7en: Given that we can actually pass multiple components I'd change this:
```suggestion
	Use:   "start component [component...]",
```

cmd/srcd/cmd/components.go
```diff
@@ -145,10 +132,71 @@ var componentsInstallCmd = &cobra.Command{
...
+			ctx := context.Background()
+			started := logAfterTimeoutWithServerLogs("this is taking a while, "+
+				"it might take a few more minutes while we install all the required images",
+				5*time.Second)
+			_, err = client.StartComponent(ctx, &api.StartComponentRequest{
+				Name: c.Name,
+			})
+			started()
+			if err != nil {
+				return humanizef(err, "could not start %s", c.Name)
```
se7entyse7en: Maybe we could build the list of errors so that we try to start each component that has been passed.

docs/commands.md
```diff
@@ -184,7 +185,17 @@ Installs source{d} Engine components images.
 
 ### srcd components start
 
-*status*: ❌ TBD
+Start source{d} Engine components.
+
+*arguments*:
+  * `component`: the name of the component image. It must be one of:
```
se7entyse7en: No mysql cli?

se7entyse7en: I don't know how much implicit it is, but maybe we should say that also the dependencies of the components are started.

cmd/srcd/cmd/components.go
smacker: `Use` of the start command matches the `Use` of install command.

cmd/srcd/cmd/components.go
smacker: not sure we really need it

docs/commands.md
smacker: updated

smacker: it's impossible to start mysql cli.

se7entyse7en: oh right!

cmd/srcd/cmd/components.go
se7entyse7en: I don't have a strong opinion on this, so it's up to you.

cmd/srcd/cmd/components.go
smacker: done

cmd/srcd/cmd/components.go
```diff
@@ -145,10 +135,71 @@ var componentsInstallCmd = &cobra.Command{
...
+		}
+
+		return nil
+	},
+}
+
+func getComponent(arg string, cmps []components.Component) (*components.Component, error) {
+	var c *components.Component
+	for _, cmp := range cmps {
+		if arg == cmp.Name || arg == cmp.Image {
```
dpordomingo: It would be so nice that we could also let the user ask for the shorter version of the component name...
it seems that `srcd-cli-` is a common and static thing, so we could also add `"srcd-cli-"+arg == cmp.Name`

smacker: yeah. I think we need this feature. But I'm not sure about this particular implementation. Maybe it's better to expose correct names in the components. I have created an issue: https://github.com/src-d/engine/issues/438

smacker:
Thanks for changelog reminder! Added

cmd/srcd/cmd/components.go
dpordomingo: I totally agree with you. Opening a separate issue was a good decision, so thanks.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
     Click to see more.
   </summary>
 
+  ### New Features
+
+  - new command `srcd components start [component]` starts source{d} component #433
```
dpordomingo: Should not be?
`srcd components start component [component...]`

carlosms:
```suggestion
  - New command `srcd components start [component]`, to start source{d} components [#433](https://github.com/src-d/engine/issues/433)
```

smacker:
rebased on master. CI is green. Merging.

PR #432, opened 2019-04-11
Fixes check in `StreamLinifier` for pending lines

se7entyse7en:
Closes #430.

PR #439, opened 2019-04-12
[WIP] Update to Gitbase v0.20.0

### LINKED ISSUE #424: Update to gitbase 0.20 beta
gitbase 0.20 is in beta. We can start to work integrating it in a new feature branch, `gitbase-v0.20.0`.

The breaking changes should not affect us. But it now includes a mysql client in the docker image, which means we can replace the new mysql image introduced in Engine v0.12.0, and use the gitbase image only.

se7entyse7en:
Closes #424.

Marked as WIP as waiting for final Gitbase v0.20.0 release.


se7entyse7en:
The tests are failing due to a bug that will be fixed by #432.

carlosms:
Looks good so far.
But it looks like there are more test that are failing, not only because of #432.

se7entyse7en:
@carlosms thanks I didn't notice. To simplify review I've just rebased to master to solve conflicts, gonna address the tests now and also update to `beta3`.

cmd/srcd/cmd/sql.go
```diff
@@ -192,81 +162,32 @@ func startGitbaseWithClient(client api.EngineClient) error {
 		return humanizef(err, "could not start gitbase")
 	}
 
-	if err := docker.EnsureInstalled(components.MysqlCli.Image, components.MysqlCli.Version); err != nil {
```
smacker: this function has name `startGitbaseWithClient` but it doesn't start client anymore. Just gitbase.

docker/docker_exec.go
```diff
@@ -0,0 +1,176 @@
...
+			var prevState *term.State
+			// set terminal into raw mode to propagate special
+			// characters
+			prevState, err := term.SetRawTerminal(fd)
+			if err != nil {
+				return err
+			}
+
+			defer func() {
+				err = term.RestoreTerminal(fd, prevState)
```
smacker: this error is never returned from the function

docker/docker_exec.go
```diff
@@ -0,0 +1,176 @@
...
+
+	insResp, err := c.ContainerExecInspect(ctx, idResp.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &insResp, nil
+}
+
+func withRawTerminal(in io.ReadCloser, fn func() error) func() error {
```
smacker: the name of this function and arguments below are a little confusing. `withRawTerminal` and `rawTerminal`. But actually you pass `true` and call this function even for non-terminal.

cmd/srcd/cmd/sql.go
```diff
@@ -192,81 +162,32 @@ func startGitbaseWithClient(client api.EngineClient) error {
...
+	if insResp.Running {
+		return fmt.Errorf("MySQL cli is still running")
 	}
-}
 
-func stopMysqlClient() {
-	err := docker.RemoveContainer(components.MysqlCli.Name)
-	if err != nil {
-		log.Warningf("could not stop mysql client: %v", err)
+	if insResp.ExitCode != 0 {
```
smacker: not sure if we have tests for it. What happens when you exit with `ctrl-c` or `ctrl-d`? Does it exit correctly with exit code 0?

se7entyse7en: nice catch! Now it exits with 0 on `ctrl-d` but not on `ctrl-c`

smacker: maybe it's okay to exit with error on `ctrl-c` in this case

cmd/srcd/cmd/sql.go
se7entyse7en: done [here](https://github.com/src-d/engine/pull/439/commits/50128fb356d9b925cd6a202479d9118bd765f772) and [here](https://github.com/src-d/engine/pull/439/commits/32b76ae48f5fe71af8df0d5c3bb286c7e2366f23).

docker/docker_exec.go
se7entyse7en: refactored this part [here](https://github.com/src-d/engine/pull/439/commits/a00769aed27ad38781378be61f758ac365fb490c). It should be clearer now and the code is more decoupled.

docker/docker_exec.go
se7entyse7en: answered [here](https://github.com/src-d/engine/pull/439#discussion_r277761137).

cmd/srcd/cmd/sql.go
se7entyse7en: added a test for `Ctrl-D` [here](https://github.com/src-d/engine/pull/439/commits/4e151e4de579f1d1bb6b0506465875debf117301).

docker/docker_exec.go
```diff
@@ -85,92 +87,91 @@ func exec(ctx context.Context, attach, interactive bool, containerName string, a
...
-
-			defer func() {
-				err = term.RestoreTerminal(fd, prevState)
-			}()
-		}
+type attachFn func(chan<- error)
 
-		return fn()
-	}
+type stdioAttaccher struct {
```
smacker: attach -> attacher

docker/docker_exec.go
```diff
@@ -85,92 +87,91 @@ func exec(ctx context.Context, attach, interactive bool, containerName string, a
...
+	var err error
+	var prevState *term.State
+	done := make(chan error)
+	wrappedDone := make(chan error)
+	if isTerminal {
+		// set terminal into raw mode to propagate special
+		// characters
+		prevState, err = term.SetRawTerminal(fd)
+		if err != nil {
+			done <- err
```
smacker: you have deadlock here

docker/docker_exec.go
se7entyse7en: 🤦‍♂️

docker/docker_exec.go
se7entyse7en: 😮 eagle's eye! thanks for noticing

docker/docker_exec.go
se7entyse7en: Addressed [here](https://github.com/src-d/engine/pull/439/commits/da01be9f2f2845ceab0eee453b137f58f0803169).

docker/docker_exec.go
se7entyse7en: Addressed [here](https://github.com/src-d/engine/pull/439/commits/c39c87b4c41dca3d963944406af51c9f1db04ab0)

docker/docker_exec.go
```diff
@@ -155,7 +155,10 @@ func (sa *stdioAttacher) withRawTerminal(fn attachFn) <-chan error {
 		// characters
 		prevState, err = term.SetRawTerminal(fd)
 		if err != nil {
-			done <- err
+			go func() {
```
smacker: maybe it's better to make the channel buffered?

se7entyse7en: Even if in this case would slightly increase readability, I think that using an unbuffered channel keeps the code simpler compared to the readability gain of using a buffered channel.

smacker: what exactly do you see as a downside of a buffered channel?

se7entyse7en: it's just an easier model to think 😄

se7entyse7en:
Rebased on `master` and updated to use `gitbase-v0.20.0-beta4`.

se7entyse7en:
Update to `gitbase-v0.20.0-rc1`.

PR #443, opened 2019-04-17
Updates CHANGELOG

se7entyse7en:
Adds CHANGELOG entry for #417.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@
 --log-force-format                     ignore if it is running on a terminal or not [$LOG_FORCE_FORMAT]
 ```
 
+### Bug Fixes
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
 
 - `bblfsh/bblfshd` has been updated to [v2.12.1-drivers](https://github.com/bblfsh/bblfshd/releases/tag/v2.12.1).
 
+### New Features
+
+- The Docker client is successfully created only if some compatibility checks pass ([#417](https://github.com/src-d/engine/issues/417)
```
carlosms: Maybe this should be a  bug fix rather than a new feature. We are solving a case where the command ended in a panic.

Also, I would provide more detail. Something along the lines of
```suggestion
- The commands fail gracefully if an incompatible Docker installation is found, such as Docker Toolbox  ([#417](https://github.com/src-d/engine/issues/417).
```

se7entyse7en: done, updated and rebased.

PR #446, opened 2019-04-17
New command srcd config

### LINKED ISSUE #422: New command to edit the config file
This issue is to have a new command `srcd config` that will open the default editor to edit the config file. If the file does not exit, it will be created, and the editor will contain the default config (https://docs.sourced.tech/engine/learn-more/commands#srcd).
This comes from a discussion in the original Design Document titled 'Engine Port Management'.

To discuss: should we have also the `--config=file` flag, to edit a config file in a path different from the default?

Let's block this until #410 is merged.

carlosms:
Fix #422.

I only knew about `EDITOR` env var. Apparently for historic reasons many programs look first for an editor set in `VISUAL`, and fall back to `EDITOR` if that one is not set.

If there isn't an editor set I chose to use `nano` for macos and linux. It should be installed by default, and if you don't have an editor preference `nano` will be more user friendly than `vim` or `emacs`.

### DIFF
--- a/cmd/srcd/cmd/config.go
+++ b/cmd/srcd/cmd/config.go
@@ -1,0 +1,98 @@
+package cmd
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/src-d/engine/cmd/srcd/config"
+	"gopkg.in/src-d/go-log.v1"
+)
+
+// configCmd represents the config command
+type configCmd struct {
+	Command `name:"config" short-description:"Edit the config file" long-description:"Edit the config file.\n\nIf the config file is empty it will be populated with the default values.\nSet EDITOR env var to change the text editor command."`
+}
+
+func (c *configCmd) Execute(args []string) error {
+	cmdName := os.Getenv("VISUAL")
+	if cmdName == "" {
+		cmdName = os.Getenv("EDITOR")
+	}
+
+	if cmdName == "" {
+		if runtime.GOOS == "windows" {
+			cmdName = "start" // opens the default editor
+		} else {
+			cmdName = "nano"
+		}
+
+		log.Debugf("EDITOR is not set, using command %s", cmdName)
+	}
+
+	configFile := c.Config
+	if configFile == "" {
+		var err error
+		configFile, err = config.DefaultPath()
+		if err != nil {
+			return err
+		}
+	}
+
+	emptyFile, err := isEmptyFile(configFile)
+	if err != nil {
+		return humanizef(err, "could not read config file contents")
+	}
+
+	if emptyFile {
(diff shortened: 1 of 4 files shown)


cmd/srcd/cmd/config.go
```diff
@@ -0,0 +1,98 @@
...
+		var err error
+		configFile, err = config.DefaultPath()
+		if err != nil {
+			return err
+		}
+	}
+
+	var emptyFile bool
+
+	_, err := os.Stat(configFile)
```
se7entyse7en: nit: I'd move all this check into a `func isEmptyConfig(path string) (bool, error)` instead of only separating part of the logic into the `fileIsEmpty` function.

cmdtests/config_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/src-d/engine/cmd/srcd/config"
+	"github.com/src-d/engine/cmdtests"
+	"github.com/stretchr/testify/suite"
+)
+
+type ConfigTestSuite struct {
```
se7entyse7en: Maybe we could also test the command without passing `--config` by overriding the `HOME` env var in `cmd.Env`.

cmd/srcd/cmd/config.go
carlosms: Done in 1c2cf99

cmdtests/config_test.go
carlosms: :+1: done in 77dadee

cmd/srcd/cmd/config.go
```diff
@@ -0,0 +1,99 @@
...
+// contains empty text
+func isEmptyFile(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return false, err
+		}
+
+		return true, nil
+	} else {
```
se7entyse7en: We can remove the `else`.

cmdtests/config_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (s *ConfigTestSuite) SetupTest() {
+	s.IntegrationTmpDirSuite.SetupTest()
+
+	// To test $HOME/.srcd/config.yml without breaking any existing installation
+	tmpHome := filepath.Join(s.TestDir, "home")
+	os.RemoveAll(tmpHome)
+	os.Setenv("HOME", tmpHome)
+
+	fmt.Println()
```
se7entyse7en: I think these are leftovers.

carlosms: :upside_down_face:  fixed in 787ab5a

cmd/srcd/cmd/config.go
carlosms: Yes, left over from copy and paste. Done in 787ab5a

PR #447, opened 2019-04-19
Fixes test regression

se7entyse7en:
I noticed that regression tests are broken. I've added a simple travis stage to at least check that these are buildable to avoid breaking them again in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,9 @@ jobs:
     - name: 'Integration Tests Linux'
       script:
         - TEST_PRUNE_WITH_IMAGE=true make test-integration
+    - name: 'Regressions Tests Buildable'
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -33,6 +33,9 @@ jobs:
     - name: 'Integration Tests Linux'
       script:
         - TEST_PRUNE_WITH_IMAGE=true make test-integration
+    - name: 'Regressions Tests Buildable'
+      script:
+        - go test -tags=regression -run=noop github.com/src-d/engine/cmdtests/
```
se7entyse7en: Taken from [here](https://github.com/golang/go/issues/15513#issuecomment-222023957).

Issue #297, opened 2019-03-06
Parsing csharp files does not recognize the language


dpordomingo:
discovered at https://github.com/src-d/empathy-sessions/issues/37

Running parse over a `.cs` file is not working

```shell
$ srcd parse uast examples/example.cs
INFO[0000] detected language: c#                        
Error: language c# is not supported
Usage:
  srcd parse uast [file-path] [flags]
...
```

smacker:
This is a known problem in bblfsh:
- [Closed Issue](https://github.com/bblfsh/client-go/issues/108)
- [Workaround that closed the issue](https://github.com/bblfsh/bblfshd/pull/251)
- [It does not include fix for supported languages](https://github.com/bblfsh/bblfshd/pull/251#pullrequestreview-203412959)
- [Issue for proper fix](https://github.com/bblfsh/bash-driver/issues/39#issuecomment-442539768)

We use [supported languages API](https://github.com/src-d/engine/blob/master/cmd/srcd/cmd/parse.go#L98) to provide meaningful error to the user.

So the question is: do we want to implement workaround inside engine specially for C#?

//cc @creachadair maybe you have some ETA for the proper fix on bblfsh side?

creachadair:
My understanding was that the fix from https://github.com/bblfsh/bblfshd/pull/251 was supposed to address this without requiring a change to Engine. To really fix this we will need a solution for aliasing, for which we don't have a timeline yet (@bzz has started an [RFC](https://groups.google.com/d/msg/bblfsh-dev/mo1o5d_CwnA/Q9G_rNiVBgAJ) for the question, but we have had higher-priority work so far).

Is the above workaround _not_ working for you?

smacker:
it does **not**. We use enry + bblfsh supported languages api to check that bblfsh is able to parse a file. This api currently doesn't return `c#`, only `csharp`.

creachadair:
> it does **not**. We use enry + bblfsh supported languages api to check that bblfsh is able to parse a file. This api currently doesn't return `c#`, only `csharp`.

I see. So is the issue not that you're unable to get a parse for the file, but that the list of languages doesn't match the string on the Engine side so that Engine doesn't even ask?

smacker:
@creachadair jfyi workaround from above also doesn't solve [the issue](https://github.com/src-d/gitbase/issues/728) for `srcd sql` and for [gemini](https://github.com/src-d/gemini/issues/218) (which isn't part of engine yet).

[...]

PR #449, opened 2019-04-19
srcd parse: do not check the list of languages

carlosms:
This is a workaround for #297.

The command now does not try to check if the file language is in the list of bblfsh drivers. But if bblfsh reports that the parsing failed because of a missing driver, engine will still print the same nice error message with the current list of supported languages.

It works for
- enry `c#` -> bblfsh `csharp`
- enry `c++` -> bblfsh `cpp`.

It still fails for
- enry `shell` -> bblfsh `bash`.

Unfortunately now that the tests parse the cpp and csharp files, they fail, so I had to disable them.
See https://github.com/bblfsh/cpp-driver/issues/31, https://github.com/bblfsh/bblfshd/issues/259

### DIFF
--- a/cmd/srcd/cmd/parse.go
+++ b/cmd/srcd/cmd/parse.go
@@ -78,8 +78,6 @@ func (cmd *parseUASTCmd) Execute(args []string) error {
 	}
 
 	lang := cmd.Lang
-	var resp *api.ListDriversResponse
-
 	if lang == "" {
 		lang, err = parseLang(ctx, c, cmd.Args.Path, b)
 		started()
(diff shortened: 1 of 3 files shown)


cmd/srcd/cmd/parse.go
```diff
@@ -123,6 +110,12 @@ func (cmd *parseUASTCmd) Execute(args []string) error {
 		}
 
 		if err != nil {
+			if strings.Contains(err.Error(), "missing driver for language") {
```
smacker: do we have an issue in bblfsh to export this error instead of `strings.Contains`?

carlosms: yes, it's this one https://github.com/src-d/engine/issues/297

smacker: wait. this is an issue in engine about aliasing.

I'm talking about an issue in bblfsh-sdk / bblfsh-go-client that would differentiate lang not found error from any other error.
I don't see any link or any conversation about it in 297.

smacker: @carlosms ping

PR #451, opened 2019-04-19
Rename 'parse drivers list' sub command to 'parse drivers'

### LINKED ISSUE #320: `srcd parse drivers` has only one sub command, it can be "moved" up one level
We removed other `srcd parse drivers` sub command for v0.8.0. Now the only one left is `srcd parse drivers list`.

I propose to remove the sub command, and make `srcd parse drivers` the command to list the available bblfsh drivers.

carlosms:
Fix #320.


carlosms:
Note to self: needs entry in changelog!

PR #452, opened 2019-04-19
Add cpu limits to gitbase container

### LINKED ISSUE #279: performance and resources releasing
Related to the empathy session.

running this query locally either on `srcd sql` or on `srcd web sql` makes my laptop got frozen
```sql
SELECT * FROM (
    SELECT cf.file_path,
           UAST(f.blob_content, LANGUAGE(f.file_path,  f.blob_content)) as uast
    FROM ref_commits r
    NATURAL JOIN commit_files cf
    NATURAL JOIN files f
    WHERE r.ref_name = 'HEAD'
    AND r.history_index = 0
) t WHERE uast != '';
```

This is the list of repositories I was using:
- github.com/srcd-/gitbase
- github.com/srcd-/gitbase-web
- github.com/bblfsh/bblfshd
- github.com/apache/spark
- github.com/spring-projects/spring-boot
- github.com/spring-projecst/spring-framework

Is this the expected behavior?

Also running a lighter query that was taking too long
```sql
SELECT repository_id, blob_content
FROM files
WHERE file_path = 'LICENSE'
OR file_path = 'README.md';
```
I noticed that after cancel the query the resources aren't release (at least properly, maybe the different services involved continue running the query) so I had to call `srcd stop`

carlosms:
Fix #279.

Queries like `select * from files order by...` can freeze the system. It's happened to me a few times, and I've had to reboot my laptop. This PR adds a limit on the container access to the host CPU.

I chose a limit of 75% a bit arbitrarily. After playing around with a few values it seemed like the sweet spot.

### DIFF
--- a/cmd/srcd-server/engine/sql.go
+++ b/cmd/srcd-server/engine/sql.go
@@ -84,7 +84,7 @@ func (s *Server) SQL(req *api.SQLRequest, stream api.Engine_SQLServer) error {
 	return errors.Wrap(rows.Err(), "closing row iterator")
 }
 
-func createGitbase(opts ...docker.ConfigOption) docker.StartFunc {
+func (s *Server) createGitbase(opts ...docker.ConfigOption) docker.StartFunc {
 	return func(ctx context.Context) error {
 		if err := docker.EnsureInstalled(gitbase.Image, gitbase.Version); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


docker/docker.go
```diff
@@ -727,3 +727,18 @@ func resizeTty(c *client.Client, containerID string) error {
 		Width:  width,
 	})
 }
+
+// NCPU returns the number of CPUs in the docker host
+func NCPU(ctx context.Context) (int, error) {
```
smacker: could you please explain/give a link what number exactly this function returns. Quick search on the internet didn't answer my question. I'm not sure how it works on linux but on macOs (and most probably windows) number of available CPUs for docker is limited by default. On my laptop with 4 cores docker uses only 2 by default.
So docker can't freeze my system (only freeze itself, but I can kill it).

Does this function return real number of CPUs or only the number available to docker?

carlosms: Right, I forgot about docker desktop, that is already using a VM with limits.
On linux it returns the number of host CPUs, 8 in my case.

So I'm think that I can modify this to limit the container CPU only if the docker host is linux. Otherwise for mac or windows you have docker desktop, and it already comes with good default limits.

smacker: I'm not aware of docker on linux in details. But if there is no default/manual limit for all docker then your solution sounds good.

carlosms: done in cb36df7

cmd/srcd-server/engine/sql.go
```diff
@@ -101,20 +101,28 @@ func createGitbase(opts ...docker.ConfigOption) docker.StartFunc {
 		}
 
 		// gitbase can be CPU intensive, set a limit of 75% of the host CPU.
-		// Docker option --cpus is not available in the api, but it is a shorthand.
-		// --cpus="1.5" is equivalent to --cpu-period="100000" --cpu-quota="150000"
 
-		ncpu, err := docker.NCPU(ctx)
+		host := &container.HostConfig{}
+		desktop, err := docker.IsDockerDesktop(ctx)
```
smacker: `Server` already has `hostOS` options.

carlosms: oh, I missed that config. Done in b5159dc, will rebase before merging.

Issue #456, opened 2019-04-24
Run integration tests on current master


carlosms:
Code is ready for 0.13 rc1, we need to run the tests on the windows dev machine.
And also manually test:
- If new log output from `go-log` looks ok
- If `srcd config` works as expected. It should be able to open the default editor, notepad.

smacker:
I have fixed some bugs & merged in master as you could see.

But we have more problems:
- `go-cli` uses `/flag` format for options instead of `--flag` as it was before, are we ok with it?
- we have changed how errors are displayed. Now they escape `\` and tests are failing
- some sql tests are failing now. Because queries like `/*_comment_*/_show_tables;` are parsed like unknown flags now. What do we want to do about it?
- `TestWebTestSuite/TestSQL` fails with ` could not prune components: unable to remove all containers: container not found` now sure why
- logs are ugly now:
<img width="1059" alt="Screenshot 2019-04-24 at 18 07 26" src="https://user-images.githubusercontent.com/406916/56675212-da5dab00-66bb-11e9-9354-1391e11a3047.png">
- looks like `StreamLinifier` doesn't support windows correctly (most probably due to `\r\n` instead of '\n'). Tests are failing.

carlosms:
> * `go-cli` uses `/flag` format for options instead of `--flag` as it was before, are we ok with it?
> * some sql tests are failing now. Because queries like `/*_comment_*/_show_tables;` are parsed like unknown flags now. What do we want to do about it?

From [jessevdk/go-flags docs](https://godoc.org/github.com/jessevdk/go-flags#hdr-Supported_features), we can change it to be `--opt` with the build tag `forceposix`.
We can change it in `run-integration-tests.bat` (`go build -tags forceposix`) and `Makefile` (`GO_TAGS = forceposix`)

> * `TestWebTestSuite/TestSQL` fails with ` could not prune components: unable to remove all containers: container not found` now sure why

This one looks like the most important problem to me.

se7entyse7en:
> - `go-cli` uses `/flag` format for options instead of `--flag` as it was before, are we ok with it?
> - some sql tests are failing now. Because queries like `/*_comment_*/_show_tables;` are parsed like unknown flags now. What do we want to do about it?

If the windows user is more familiar with `/` instead of `-` I'd keep this behavior. WRT the tests I think that it makes sense to have different test builds for Windows. In this `/*_comment_*/_show_tables;` case on windows is actually expected that it raises an unknown flags error.

[...]

PR #461, opened 2019-04-26
Bump gopkg.in/src-d/go-log.v1 to v1.0.2

carlosms:
Part of #456.
Includes the fix for wrong log color characters in windows.


carlosms:
After testing the bin in windows, the logs still show the same problem. #462.

